PR #21, opened 2018-08-13
Modify resources

DGB2018:
Modify resources feefines/ and accounts/

### DIFF
--- a/ramls/accountdata.json
+++ b/ramls/accountdata.json
@@ -81,19 +81,7 @@
     "loanId": {
       "type": "string"
     },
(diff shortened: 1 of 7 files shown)


mreno-EBSCO:
I know this is already merged, but can we put back the "itemId" field in accountdata? mod-patron uses this field to lookup info about the item the fee/fine is for and without it we'd have to do a barcode search, which, as it currently stands, is a **very** slow query. Lookup by ID is a much preferred approach, I'm curious what the reason is that we need to remove all those IDs?

PR #48, opened 2019-08-09
Update RMB 26.2.4

DGB2018:
* Update 26.2.4 RMB version. [Version 25](https://github.com/folio-org/raml-module-builder/blob/master/doc/upgrading.md#version-25)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "raml-util"]
 	path = raml-util
 	url = https://github.com/folio-org/raml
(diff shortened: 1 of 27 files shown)


dcrossleyau:
Glancing at the earlier commits in this branch, i see old files from the "v0.8" raml-util. So perhaps your git submodule has become confused.

dcrossleyau:
You now have an extra erroneous entry in your .gitmodules config as "path = raml-util". Need to remove that configuration entry leaving only the "path = ramls/raml-util" entry.

PR #49, opened 2019-08-15
[MODFEE-11 MODFEE-12] Send fee/fine patron notices

opochepam:
## Purpose
Resolves: MODFEE-11, MODFEE-12
1. Send patron notice for payment of fee/fine charge
2. Send patron notice for manual fee/fine charge
## Links
https://issues.folio.org/browse/MODFEE-11
https://issues.folio.org/browse/MODFEE-12

### DIFF
--- a/ramls/accounts.raml
+++ b/ramls/accounts.raml
@@ -11,6 +11,7 @@ types:
   account: !include accountdata.json
   accountdataCollection: !include accountdataCollection.json
   errors: !include raml-util/schemas/errors.schema
+  patronNotice: !include patronNotice.json
 
 traits:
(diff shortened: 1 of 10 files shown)


ramls/accounts.raml
```diff
@@ -11,6 +11,7 @@ types:
   account: !include accountdata.json
   accountdataCollection: !include accountdataCollection.json
   errors: !include raml-util/schemas/errors.schema
+  patronNotice: !include patronNotice.json
```
kostyantyn-kh: Minor version here and in module-descriptor should be increased because of this change

PR #51, opened 2019-09-03
[MODFEE-13] Add fee/fine tokens to patron notices

dmytrotur:
## Purpose
Add fee/fine tokens to patron notices
Resolves: [MODFEE-13](https://issues.folio.org/browse/MODFEE-13)


src/main/java/org/folio/rest/domain/FeeFineNoticeContext.java
```diff
@@ -0,0 +1,81 @@
...
+
+public class FeeFineNoticeContext {
+
+  private static final String PAID_FULLY = "Paid fully";
+  private static final String PAID_PARTIALLY = "Paid partially";
+
+  private Owner owner;
+  private Feefine feefine;
+  private Account account;
+  private Feefineaction feefineaction;
```
kostyantyn-kh: Could this be renamed to FeeFineAction?

src/main/java/org/folio/rest/domain/FeeFineNoticeContext.java
```diff
@@ -0,0 +1,81 @@
...
+    this.feefine = feefine;
+    this.account = account;
+    this.feefineaction = feefineaction;
+  }
+
+  public String getTemplateId() {
+    String templateId = null;
+
+    if (feefineaction.getPaymentMethod() == null) {
+      templateId = feefine.getChargeNoticeId() != null ?
```
kostyantyn-kh: Could be simplified: return feefine...

src/main/java/org/folio/rest/repository/OwnerRepository.java
```diff
@@ -15,9 +16,9 @@ public OwnerRepository(PostgresClient pgClient) {
     this.pgClient = pgClient;
   }
 
-  public Future<Owner> getById(String id) {
+  public Future<FeeFineNoticeContext> loadOwner(FeeFineNoticeContext context) {
     Future<Owner> future = Future.future();
-    pgClient.getById(OWNERS_TABLE, id, Owner.class, future);
-    return future;
+    pgClient.getById(OWNERS_TABLE, context.getFeefine().getOwnerId(), Owner.class, future);
```
kostyantyn-kh: Probably extra check for null is needed here (same for other repository files)

src/main/java/org/folio/rest/domain/FeeFineNoticeContext.java
dmytrotur: Done

src/main/java/org/folio/rest/repository/OwnerRepository.java
dmytrotur: Done

PR #59, opened 2019-11-06
[MODFEE-16] Error message not displaying when duplicate policy name entered

OleksandrVidinieiev:
**Purpose**
Fix a bug in `mod-feesfines`

**Link**
[MODFEE-16](https://issues.folio.org/browse/MODFEE-16)

### DIFF
--- a/src/main/java/org/folio/rest/impl/OverdueFinePoliciesAPI.java
+++ b/src/main/java/org/folio/rest/impl/OverdueFinePoliciesAPI.java
@@ -48,7 +48,6 @@ public class OverdueFinePoliciesAPI implements OverdueFinesPolicies {
             Handler<AsyncResult<Response>> asyncResultHandler,
             Context vertxContext) {
 
-
     PgUtil.post(OVERDUE_FINE_POLICY_TABLE, entity, okapiHeaders, vertxContext, PostOverdueFinesPoliciesResponse.class,
       r -> {
         Response response = r.result();
--- a/src/main/java/org/folio/rest/impl/OverdueFinePoliciesAPI.java
+++ b/src/main/java/org/folio/rest/impl/OverdueFinePoliciesAPI.java
@@ -2,20 +2,28 @@ package org.folio.rest.impl;
 
 import io.vertx.core.AsyncResult;
 import io.vertx.core.Context;
+import io.vertx.core.Future;
 import io.vertx.core.Handler;
-import java.util.Map;
-import javax.ws.rs.core.Response;
 import org.folio.rest.annotations.Validate;
+import org.folio.rest.jaxrs.model.Errors;
+import org.folio.rest.jaxrs.model.Error;
 import org.folio.rest.jaxrs.model.OverdueFinePolicies;
 import org.folio.rest.jaxrs.model.OverdueFinePolicy;
 import org.folio.rest.jaxrs.model.OverdueFinesPoliciesGetOrder;
 import org.folio.rest.jaxrs.resource.OverdueFinesPolicies;
 import org.folio.rest.persist.PgUtil;
+import org.folio.rest.utils.ErrorHelper;
+
+import javax.ws.rs.core.Response;
+import java.util.Collections;
+import java.util.Map;
 
 public class OverdueFinePoliciesAPI implements OverdueFinesPolicies {
 
-    public static final String OVERDUE_FINE_POLICY_TABLE = "overdue_fine_policy";
-    private static final Class<OverdueFinePolicy> OVERDUE_FINE_POLICY = OverdueFinePolicy.class;
+  static final String OVERDUE_FINE_POLICY_TABLE = "overdue_fine_policy";
+  static final String DUPLICATE_ERROR_CODE = "feesfines.policy.overdue.duplicate";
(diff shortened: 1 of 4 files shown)


src/main/java/org/folio/rest/utils/ErrorHelper.java
```diff
@@ -0,0 +1,42 @@
...
+    p.setValue(value);
+    error.getParameters().add(p);
+    error.setMessage(message);
+    error.setCode("-1");
+    error.setType(RTFConsts.VALIDATION_FIELD_ERROR);
+    List<Error> l = new ArrayList<>();
+    l.add(error);
+    e.setErrors(l);
+    return e;
+  }
```
OleksiiMaksymov: The method is used only for test cases. This method should be moved to the test package.
It can also be simplified:

  private Errors createValidationErrorMessage(String field, String value, String message) {

    Parameter parameter = new Parameter()
      .withKey(field)
      .withValue(value);

    Error error = new Error()
      .withParameters(Collections.singletonList(parameter))
      .withCode("unique code")
      .withMessage(message)
      .withType(RTFConsts.VALIDATION_FIELD_ERROR);

    return new Errors()
      .withErrors(Collections.singletonList(error));
  }

src/test/java/org/folio/rest/utils/ErrorHelperTest.java
```diff
@@ -0,0 +1,45 @@
...
+    Errors message = ErrorHelper.createValidationErrorMessage("field", "value", "message");
+    assertNotNull(message);
+    assertEquals(1, message.getErrors().size());
+
+    Error error = message.getErrors().get(0);
+    assertEquals("message", error.getMessage());
+
+    assertNotNull(error.getParameters());
+    assertEquals(1, error.getParameters().size());
+
```
OleksiiMaksymov: The test skipped code verification:
`assertEquals("unique code", error.getCode());`

src/main/java/org/folio/rest/utils/ErrorHelper.java
```diff
@@ -0,0 +1,42 @@
...
+
+  public static boolean didUniqueConstraintViolationOccur(Response response, String constraintName) {
+    return response != null &&
+      response.hasEntity() &&
+      response.getEntity() instanceof String &&
+      ((String) response.getEntity())
+        .startsWith(("duplicate key value violates unique constraint \"" + constraintName + "\""));
```
OleksiiMaksymov: Magic variable:
`private static final String ERROR_MSG = "duplicate key value violates unique constraint \"%s\"";`

src/main/java/org/folio/rest/utils/ErrorHelper.java
```diff
@@ -0,0 +1,42 @@
...
+    l.add(error);
+    e.setErrors(l);
+    return e;
+  }
+
+  public static boolean didUniqueConstraintViolationOccur(Response response, String constraintName) {
+    return response != null &&
+      response.hasEntity() &&
+      response.getEntity() instanceof String &&
+      ((String) response.getEntity())
```
OleksiiMaksymov: Readability can be improved using an Optional:
```
 return Optional.ofNullable(response)
      .filter(Response::hasEntity)
      .map(Response::getEntity)
      .filter(r -> r instanceof String)
      .filter(r -> ((String) r).contains(String.format(ERROR_MSG, constraintName)))
      .isPresent();
```

src/main/java/org/folio/rest/impl/OverdueFinePoliciesAPI.java
```diff
@@ -1,83 +1,103 @@
...
+                    Response response = r.result();
+                    if (ErrorHelper.didUniqueConstraintViolationOccur(response, PRIMARY_KEY)) {
+                        Error error = new Error()
+                                .withMessage(DUPLICATE_ENTITY_MESSAGE)
+                                .withCode(DUPLICATE_ERROR_CODE);
+                        Errors errors = new Errors().withErrors(Collections.singletonList(error));
+                        response = PostOverdueFinesPoliciesResponse.respond422WithApplicationJson(errors);
+                    }
+                    asyncResultHandler.handle(Future.succeededFuture(response));
+                });
```
OleksiiMaksymov: here you can also use: r.map(....)

    public void postOverdueFinesPolicies(......) {

      PgUtil.post(OVERDUE_FINE_POLICY_TABLE, entity, okapiHeaders, context,
        PostOverdueFinesPoliciesResponse.class, r -> {
          if (ErrorHelper.didUniqueConstraintViolationOccur(r.result(), PRIMARY_KEY)) {
            Error error = new Error()
              .withMessage(DUPLICATE_ENTITY_MESSAGE)
              .withCode(DUPLICATE_ERROR_CODE);
            asyncResultHandler.handle(r.map(PostOverdueFinesPoliciesResponse.respond422WithApplicationJson(createErrors(error))));
            return;
          }
          asyncResultHandler.handle(r);
        });
    }

OleksiiMaksymov: private Errors createErrors(Error ... errors){
      return new Errors()
        .withErrors(Arrays.asList(errors));
    }

PR #62, opened 2019-11-08
[MODFEE-17] Error message not displaying when duplicate policy name entered

OleksandrVidinieiev:
**Purpose**
Fix a bug in `mod-feesfines`.

**Link**
[MODFEE-17](https://issues.folio.org/browse/MODFEE-17)

### DIFF
--- a/src/test/java/org/folio/rest/impl/LostItemFeePoliciesAPITest.java
+++ b/src/test/java/org/folio/rest/impl/LostItemFeePoliciesAPITest.java
@@ -125,7 +125,7 @@ public class LostItemFeePoliciesAPITest {
         post(lostItemFeePolicyEntity);
 
         JsonObject errorJson = new JsonObject()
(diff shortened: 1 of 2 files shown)


src/test/java/org/folio/rest/impl/LostItemFeePoliciesAPITest.java
```diff
@@ -0,0 +1,249 @@
...
+
+    @BeforeClass
+    public static void setUpClass(final TestContext context) throws Exception {
+        Async async = context.async();
+        vertx = Vertx.vertx();
+        port = NetworkUtils.nextFreePort();
+        lostItemFeePolicyEntity = createEntity();
+
+        PostgresClient.setIsEmbedded(true);
+        PostgresClient.getInstance(vertx).startEmbeddedPostgres();
```
kostyantyn-kh: startEmbeddedPostgres method already has call to setEmbedded in it

PR #67, opened 2019-12-26
MODFEE-22 Sending "new fee/fine" and "payment" notices when "Notify patron" unchecked

Svitlana-Zmiivska:
**Purpose**
Fix a bug in `mod-feesfines`

**Link**
[MODFEE-22](https://issues.folio.org/browse/MODFEE-22)

### DIFF
--- a/src/test/java/org/folio/rest/impl/FeeFineActionsAPITest.java
+++ b/src/test/java/org/folio/rest/impl/FeeFineActionsAPITest.java
@@ -1,0 +1,313 @@
+package org.folio.rest.impl;
+
+import static com.github.tomakehurst.wiremock.client.WireMock.equalToJson;
+import static com.github.tomakehurst.wiremock.client.WireMock.matching;
+import static com.github.tomakehurst.wiremock.client.WireMock.ok;
+import static com.github.tomakehurst.wiremock.client.WireMock.postRequestedFor;
+import static com.github.tomakehurst.wiremock.client.WireMock.urlPathEqualTo;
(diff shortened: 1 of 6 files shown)


src/test/java/org/folio/rest/impl/FeeFineActionsAPITest.java
```diff
@@ -0,0 +1,313 @@
...
+  public static void tearDownClass(final TestContext context) {
+    Async async = context.async();
+    vertx.close(context.asyncAssertSuccess(res -> {
+      PostgresClient.stopEmbeddedPostgres();
+      async.complete();
+    }));
+  }
+
+  @Test
+  public void postFeefineactionsWithPatronNotification() {
```
OleksandrVidinieiev:
```suggestion
  public void postFeeFineActionsWithPatronNotification() {
```

src/test/java/org/folio/rest/impl/FeeFineActionsAPITest.java
```diff
@@ -0,0 +1,313 @@
...
+    Awaitility.await()
+      .atMost(5, TimeUnit.SECONDS)
+      .untilAsserted(() -> userMockServer.verify(postRequestedFor(urlPathEqualTo("/patron-notice"))
+        .withRequestBody(equalToJson(expectedNoticeJson))
+      ));
+
+  }
+
+  @Test
+  public void postFeefineactionsWithoutPatron() {
```
OleksandrVidinieiev:
```suggestion
  public void postFeeFineActionsWithoutPatronNotification() {
```

PR #68, opened 2020-02-07
MODFEE-23: Add sample overdue-fine and lost-item-fee policies

bohdan-suprun:
https://issues.folio.org/browse/MODFEE-23: Create Overdue Fine Policy and Lost Item Policy - Test data for snapshot/testing environments.

* Add sample overdue and lost-item policies to use them in circulation rules.
* Add tenant API.

### DIFF
--- a/src/test/java/org/folio/rest/impl/TenantRefAPITest.java
+++ b/src/test/java/org/folio/rest/impl/TenantRefAPITest.java
@@ -1,8 +1,12 @@
 package org.folio.rest.impl;
 
 import static io.vertx.core.Future.succeededFuture;
+import static org.folio.rest.RestVerticle.OKAPI_HEADER_TENANT;
+import static org.folio.rest.RestVerticle.OKAPI_HEADER_TOKEN;
 
 import java.util.Collections;
+
+import javax.ws.rs.core.MediaType;
 
 import org.folio.rest.RestVerticle;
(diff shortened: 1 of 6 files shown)


bohdan-suprun:
> I imagine most of the production code is taken from another module. 

Yes, the code taken from circulation-storage.

Svitlana-Zmiivska:
@marcjohnson-kint @bohdan-suprun As I remember Vega team is also not responsible for this module

src/test/java/org/folio/rest/impl/TenantRefAPITest.java
```diff
@@ -0,0 +1,146 @@
...
+    PostgresClient.getInstance(vertx).startEmbeddedPostgres();
+    final TenantClient tenantClient = createTenantClient();
+
+    DeploymentOptions restDeploymentOptions = new DeploymentOptions()
+      .setConfig(new JsonObject().put("http.port", PORT));
+
+    vertx.deployVerticle(RestVerticle.class.getName(), restDeploymentOptions,
+      res -> {
+        try {
+          tenantClient.postTenant(getTenantAttributes(), result -> async.complete());
```
julianladisch: `result -> async.complete()` should be replaced by `context.asyncAssertSuccess(result -> async.complete())` to fail the test when postTenant fails.

PR #72, opened 2020-02-13
MODFEE-19 Implement waive notices, add tests

alexanderkurash:
Resolves [MODFEE-19](https://issues.folio.org/browse/MODFEE-19):
- Send patron notice when fee/fine charge is waived. 
- Rename template token `fee.actionAdditionalInfo` -> `fee.actionAdditionalInfoForPatron`.

### DIFF
--- a/src/main/java/org/folio/rest/domain/FeeFineNoticeContext.java
+++ b/src/main/java/org/folio/rest/domain/FeeFineNoticeContext.java
@@ -9,6 +9,8 @@ public class FeeFineNoticeContext {
 
   private static final String PAID_FULLY = "Paid fully";
   private static final String PAID_PARTIALLY = "Paid partially";
+  private static final String WAIVED_FULLY = "Waived fully";
+  private static final String WAIVED_PARTIALLY = "Waived partially";
 
   private Owner owner;
   private Feefine feefine;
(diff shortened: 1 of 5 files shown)


marcjohnson-kint:
@alexanderkurash Please can you help me understand why you've requested a review from me for this change? Are you only looking for my feedback?

(mod-feefines does not have a review policy, so no review is required)

alexanderkurash:
> @alexanderkurash Please can you help me understand why you've requested a review from me for this change? Are you only looking for my feedback?
> 
> (mod-feefines does not have a review policy, so no review is required)

I've seen you approving one of the previous PRs. But you're right - if there's no review policy we'll handle it within Vega team.

marcjohnson-kint:
> > @alexanderkurash Please can you help me understand why you've requested a review from me for this change? Are you only looking for my feedback?
> > (mod-feefines does not have a review policy, so no review is required)
> 
> I've seen you approving one of the previous PRs. But you're right - if there's no review policy we'll handle it within Vega team.

That was only because someone asked me to specifically (which I'm happy to do). I only wanted to clarify that it isn't necessary in this case.

src/main/java/org/folio/rest/domain/FeeFineNoticeContext.java
```diff
@@ -44,7 +46,9 @@ private boolean isCharge() {
 
   private boolean isFeeFineActon() {
     return feefineaction.getTypeAction().equals(PAID_FULLY) ||
-      feefineaction.getTypeAction().equals(PAID_PARTIALLY);
+      feefineaction.getTypeAction().equals(PAID_PARTIALLY) ||
+      feefineaction.getTypeAction().equals(WAIVED_FULLY) ||
+      feefineaction.getTypeAction().equals(WAIVED_PARTIALLY);
```
OleksandrVidinieiev: Maybe we should introduce a constant list of all fee/fine actions and replace this method's body with simple `List#contains` invocation? Keep in mind that this list is only going to grow in subsequent tasks.

alexanderkurash: Done

PR #75, opened 2020-02-26
[MODFEE-34] Can't save Overdue fine policy

OleksandrVidinieiev:
**Purpose**
Fix `OverdueFinePolicy` validation for entities with fine amounts < 1.

**Approach**
Remove minimum value constraint from JSON-schema.

This resolves [MODFEE-34](https://issues.folio.org/browse/MODFEE-34).

### DIFF
--- a/ramls/quantity.json
+++ b/ramls/quantity.json
@@ -4,9 +4,8 @@
   "type": "object",
   "properties": {
     "quantity": {
-      "description": "Quantity of the number of times the interval repeats",
-      "type": "number",
-      "minimum": 1.0
+      "description": "Recurring fine amount",
+      "type": "number"
     },
     "intervalId": {
       "description": "Interval for the period, e.g. hour, day, week, month or year",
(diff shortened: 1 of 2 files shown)


ramls/quantity.json
```diff
@@ -4,9 +4,8 @@
   "type": "object",
   "properties": {
     "quantity": {
-      "description": "Quantity of the number of times the interval repeats",
-      "type": "number",
-      "minimum": 1.0
+      "description": "Recurring fine amount",
```
alexanderkurash: To avoid confusion in the future, it's probably better to emphasize on the relationship between quantity and interval in the description, something like "Fine amount per interval"

alexanderkurash:
In Jira comments Zak mentions that "the API should return a 400 with a useful error message that can be presented in the UI". What will happen if we try to create an overdue fine policy with quantity below zero?

OleksandrVidinieiev:
> In Jira comments Zak mentions that "the API should return a 400 with a useful error message that can be presented in the UI". What will happen if we try to create an overdue fine policy with quantity below zero?

There is a validation on the UI that will not let you save a value below zero.

ramls/quantity.json
OleksandrVidinieiev: done

PR #76, opened 2020-03-03
[MODFEE-28]Create automatic FeeFine records on module initialization

VitaliiZahoruiko:


src/main/java/org/folio/rest/utils/AutomaticFeeFineInitializer.java
```diff
@@ -0,0 +1,62 @@
...
+import io.vertx.core.Handler;
+
+import static java.util.Arrays.asList;
+
+public class AutomaticFeeFineInitializer {
+
+  private static final String FEEFINES_TABLE_NAME = "feefines";
+  private PostgresClient postgresClient;
+  private Handler<AsyncResult<String>> replyHandler;
+  private List<String> feefines;
```
alexanderkurash: Could any of these be `final`? It seems like `postgresClient` and `feefines` are never modified. Can `replyHandler` be passed to a constructor as well rather than set with a setter method?

src/main/java/org/folio/rest/utils/AutomaticFeeFineInitializer.java
```diff
@@ -0,0 +1,62 @@
...
+
+  private Metadata generateMetadata() {
+    Metadata metadata = new Metadata();
+    metadata.setCreatedDate(new Date());
+    return metadata;
+  }
+
+  public void setPostgresClient(PostgresClient postgresClient) {
+    this.postgresClient = postgresClient;
+  }
```
alexanderkurash: Are we really using this setter for `postgresClient`?

src/main/java/org/folio/rest/utils/AutomaticFeeFineInitializer.java
```diff
@@ -0,0 +1,62 @@
...
+
+  public void setReplyHandler(Handler<AsyncResult<String>> replyHandler) {
+    this.replyHandler = replyHandler;
+  }
+
+  public PostgresClient getPostgresClient() {
+    return postgresClient;
+  }
```
alexanderkurash: Same question about the getter.

src/test/java/org/folio/PatronNoticeTest.java
```diff
@@ -89,8 +89,8 @@ void manualChargeNoticeShouldBeSentWithDefaultTemplate() {
     String accountId = UUID.randomUUID().toString();
     String defaultChargeTemplateId = UUID.randomUUID().toString();
     String userId = UUID.randomUUID().toString();
-    String feeFineType = "damaged book";
-    String typeAction = "damaged book";
+    String feeFineType = "damaged book1";
+    String typeAction = "damaged book1";
```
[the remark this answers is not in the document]
VitaliiZahoruiko: We need different names for feefines because now names cannot be duplicated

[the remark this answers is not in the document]
alexanderkurash: Oh, I see. But these `feefine` objects are created in different tests, does it mean they're not isolated from each other? Maybe it makes sense to clean up this table before each test?

ramls/feefinedata.json
```diff
@@ -37,6 +42,7 @@
   "additionalProperties": false,
   "required": [
     "feeFineType",
-    "ownerId"
+    "ownerId",
+    "automatic"
```
OleksandrVidinieiev: I don't think FE adopted the use of this new flag yet, so manual fines are created without it at the moment. So maybe we shouldn't make it required now.

PR #80, opened 2020-03-18
[MODFEE-36] Update RMB to v29.3.1

OleksandrVidinieiev:
Update RMB to v29.3.1, update Vert.x to v3.8.5

Resolves [MODFEE-36](https://issues.folio.org/browse/MODFEE-36).

### DIFF
--- a/src/main/java/org/folio/rest/utils/ErrorHelper.java
+++ b/src/main/java/org/folio/rest/utils/ErrorHelper.java
@@ -9,23 +9,36 @@ import org.folio.rest.jaxrs.model.Error;
 import org.folio.rest.jaxrs.model.Errors;
 
 public class ErrorHelper {
-
-  private static final String UNIQUE_CONSTRAINT_MSG_TEMPLATE = "duplicate key value violates unique constraint \"%s\"";
+  private static final String DUPLICATE_NAME_MSG_TEMPLATE =
+    "lower(f_unaccent(jsonb ->> 'name'::text)) value already exists in table %s";
 
   private ErrorHelper() {
   }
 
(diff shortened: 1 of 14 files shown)


src/test/java/org/folio/rest/impl/LostItemFeePoliciesAPITest.java
```diff
@@ -41,207 +43,223 @@
 
 @RunWith(VertxUnitRunner.class)
 public class LostItemFeePoliciesAPITest {
```
alexanderkurash: Indentation is not the only change in this class, is that correct?

src/main/java/org/folio/rest/utils/ErrorHelper.java
```diff
@@ -9,23 +9,36 @@
...
+      .filter(msg -> msg.contains(String.format(DUPLICATE_NAME_MSG_TEMPLATE, tableName)))
       .isPresent();
   }
 
   public static Errors createErrors(Error... errors){
     return new Errors()
       .withErrors(Arrays.asList(errors));
   }
+
+  public static Errors createSingleError(String message, String code) {
```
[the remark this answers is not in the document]
OleksandrVidinieiev: Done.

src/test/java/org/folio/rest/impl/LostItemFeePoliciesAPITest.java
OleksandrVidinieiev: Yes. I brought back the original indentation.

alexanderkurash: Thanks, it makes it much easier to review. We'll fix indentation some time later with a separate PR.

PR #85, opened 2020-04-03
MODFEE-48 Add new fields fields to account

alexanderkurash:
Adding two fields - holdingsRecordId and instanceId. The are populated on the fly for `/accounts` and `/account/{id}` requests.

This resolves MODFEE-48 which in turn helps to resolve UIU-1553 bug.

### DIFF
--- a/src/main/java/org/folio/rest/client/InventoryClient.java
+++ b/src/main/java/org/folio/rest/client/InventoryClient.java
@@ -1,0 +1,109 @@
+package org.folio.rest.client;
+
+import static io.vertx.core.Future.failedFuture;
+import static io.vertx.core.Future.succeededFuture;
+import static javax.ws.rs.core.HttpHeaders.ACCEPT;
+import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import org.folio.rest.jaxrs.model.HoldingsRecords;
+import org.folio.rest.jaxrs.model.Items;
+
+import io.vertx.core.Future;
+import io.vertx.core.Promise;
+import io.vertx.core.buffer.Buffer;
+import io.vertx.ext.web.client.HttpRequest;
+import io.vertx.ext.web.client.HttpResponse;
+import io.vertx.ext.web.client.WebClient;
+
+public class InventoryClient extends OkapiClient {
+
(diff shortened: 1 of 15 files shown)


src/main/java/org/folio/rest/client/InventoryClient.java
```diff
@@ -0,0 +1,109 @@
...
+
+    HttpRequest<Buffer> request = okapiGetAbs("/inventory/items");
+    if (!itemIds.isEmpty()) {
+      request.addQueryParam("query", String.format("(id==(%s))",
+        itemIds.stream()
+          .map(id -> String.format("\"%s\"", id))
+          .collect(Collectors.joining(" or "))))
+      .addQueryParam("limit", ITEMS_LIMIT);
+    }
+    request.send(promise);
```
OleksandrVidinieiev, line 50: We should create a tech debt issue about splitting this call into multiple calls in case the length of the URL exceeds server's limit.

src/main/java/org/folio/rest/client/InventoryClient.java
```diff
@@ -0,0 +1,109 @@
...
+    HttpRequest<Buffer> request = okapiGetAbs("/holdings-storage/holdings");
+    if (!holdingIds.isEmpty()) {
+      request.addQueryParam("query", String.format("(id==(%s))",
+        holdingIds.stream()
+          .map(id -> String.format("\"%s\"", id))
+          .collect(Collectors.joining(" or "))))
+        .addQueryParam("limit", HOLDINGS_LIMIT);
+    }
+    request.putHeader(ACCEPT, APPLICATION_JSON);
+    request.send(promise);
```
OleksandrVidinieiev, line 88: We should create a tech debt issue about splitting this call into multiple calls in case the length of the URL exceeds server's limit.

src/main/java/org/folio/rest/impl/AccountsAPI.java
```diff
@@ -248,4 +308,22 @@ public void putAccountsByAccountId(String accountId,
       PgUtil.put(ACCOUNTS_TABLE, entity, accountId, okapiHeaders, vertxContext,
         PutAccountsByAccountIdResponse.class, asyncResultHandler);
     }
+
+    private class AdditionalFieldsContext {
+      Items items;
+      HoldingsRecords holdings;
```
OleksandrVidinieiev: Should we make them final?

alexanderkurash: Done

PR #87, opened 2020-04-07
MODFEE-50 Remove readonly fields validation

alexanderkurash:
Resolves [MODFEE-50](https://issues.folio.org/browse/MODFEE-50).

FE is trying to update these fields with `PUT /accounts/{id}` request, passing empty
strings. Though these fields are readonly and should be ignored by RMB, it still validates them which causes validation failures.

### DIFF
--- a/ramls/accountdata.json
+++ b/ramls/accountdata.json
@@ -135,14 +135,12 @@
     "holdingsRecordId": {
       "description": "Item field: item.holdingsRecordId",
       "type": "string",
-      "readonly": true,
-      "$ref": "uuid.json"
+      "readonly": true
     },
     "instanceId": {
       "description": "Holdings record field: holdingsRecord.instanceId",
       "type": "string",
-      "readonly": true,
-      "$ref": "uuid.json"
+      "readonly": true
     }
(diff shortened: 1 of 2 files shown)


ramls/accountdata.json
```diff
@@ -135,14 +135,12 @@
     "holdingsRecordId": {
       "description": "Item field: item.holdingsRecordId",
       "type": "string",
-      "readonly": true,
-      "$ref": "uuid.json"
+      "readonly": true
```
OleksandrVidinieiev, line 138: This way we are not validating UUID when it **is** present.

src/test/java/org/folio/rest/impl/AccountsAPITest.java
```diff
@@ -126,7 +125,7 @@ public void canGetAccounts() {
       .statusCode(HttpStatus.SC_CREATED)
       .contentType(ContentType.JSON);
 
-    get("/accounts")
+    get("")
```
OleksandrVidinieiev, line 128: Personally, I'd leave the absolute path

PR #88, opened 2020-05-08
MODFEE-53] Publish fee/fine balance change event

OleksandrVidinieiev:
**Purpose**
Introduce integration with mod-pubsub. Publish events when account is created, changed or deleted.

**Resolves**
[MODFEE-53](https://issues.folio.org/browse/MODFEE-53)

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -13,6 +13,22 @@
     {
       "id": "holdings-storage",
       "version": "4.2"
+    },
+    {
+      "id": "pubsub-event-types",
+      "version": "0.1"
+    },
+    {
+      "id": "pubsub-publishers",
+      "version": "0.1"
+    },
+    {
+      "id": "pubsub-subscribers",
+      "version": "0.1"
+    },
+    {
+      "id": "pubsub-publish",
+      "version": "0.1"
     }
   ],
   "provides":[
@@ -137,7 +153,8 @@
             "accounts.item.post"
           ],
           "modulePermissions": [
-            "patron-notice.post"
+            "patron-notice.post",
+            "pubsub.publish.post"
           ]
         },
         {
@@ -149,7 +166,8 @@
             "accounts.item.put"
           ],
           "modulePermissions": [
-            "patron-notice.post"
+            "patron-notice.post",
+            "pubsub.publish.post"
           ]
         },
         {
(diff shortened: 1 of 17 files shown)


src/test/java/org/folio/rest/impl/APITests.java
```diff
@@ -0,0 +1,114 @@
...
+
+import io.vertx.core.DeploymentOptions;
+import io.vertx.core.Vertx;
+import io.vertx.core.json.JsonObject;
+import io.vertx.core.logging.Logger;
+import io.vertx.core.logging.LoggerFactory;
+import io.vertx.ext.unit.Async;
+import io.vertx.ext.unit.TestContext;
+
+public abstract class APITests {
```
roman-barannyk, line 34: It seems like this class doesn't have any abstract methods. Why is it abstract?

descriptors/ModuleDescriptor-template.json
```diff
@@ -620,7 +620,12 @@
...
         {
           "methods": ["POST"],
-          "pathPattern": "/_/tenant"
+          "pathPattern": "/_/tenant",
+          "modulePermissions": [
+            "pubsub.event-types.post",
+            "pubsub.publishers.post",
+            "pubsub.subscribers.post"
+          ]
         }, {
```
alexanderkurash, line 629: Please add PubSub interfaces dependencies and permissions

src/test/java/org/folio/rest/impl/APITests.java
```diff
@@ -0,0 +1,113 @@
...
+  protected static final String OKAPI_TENANT = "test_tenant";
+  protected static final String OKAPI_TOKEN = buildToken();
+  protected static final String OKAPI_HEADER_URL = "x-okapi-url";
+
+  protected static Vertx vertx;
+  protected static OkapiClient okapiClient;
+  protected static TenantClient tenantClient;
+
+  @Rule
+  public WireMockRule wireMock = new WireMockRule(
```
alexanderkurash, line 51: Move wireMock to the class where it's being used

descriptors/ModuleDescriptor-template.json
OleksandrVidinieiev: added

src/test/java/org/folio/rest/impl/APITests.java
OleksandrVidinieiev: done

PR #89, opened 2020-05-14
MODFEE-54: Publish account with loan closed event.

bohdan-suprun:
https://issues.folio.org/browse/MODFEE-54 - Publish event when an account with loan is closed.

## Changes 

* Publish `LOAN_RELATED_FEE_FINE_CLOSED` when an account with no remaining amount and a loan associated is closed;
* Add dependency on pub/sub APIs;
* Refactor Account and TenantRefApi tests;

Design: https://wiki.folio.org/display/DD/Close+loan+when+lost+item+fees+are+closed#Closeloanwhenlostitemfeesareclosed-6Usepub/subapproachoverview


OleksandrVidinieiev:
Hi @bohdan-suprun, check out https://github.com/folio-org/mod-feesfines/pull/88. I think we may have duplicated a ton of work :(

bohdan-suprun:
> Hi @bohdan-suprun, check out #88. I think we may have duplicated a ton of work :(

Wow, you're right @OleksandrVidinieiev. Thanks for pointing to this.

OleksandrVidinieiev:
> 
> 
> > Hi @bohdan-suprun, check out #88. I think we may have duplicated a ton of work :(
> 
> Wow, you're right @OleksandrVidinieiev. Thanks for pointing to this.

I will review this PR tomorrow. Maybe your implementation is better :)

OleksandrVidinieiev:
> 
> 
> > > Hi @bohdan-suprun, check out #88. I think we may have duplicated a ton of work :(
> > 
> > 
> > Wow, you're right @OleksandrVidinieiev. Thanks for pointing to this.
> 
> I will review this PR tomorrow. Maybe your implementation is better, so don't try to merge my changes into your branch just yet :)

I looked through this PR, and yes, we did have a lot of the same ideas, like extraction of a base class for API tests, introduction of OkapiClient, and even that method for token generation:) One thing I failed at though is mocking the calls to `PubSubClient` with WireMock: I kept getting the same error about a missing implementation of interface `Pubsub` found in `mod-pubsub-client` jar. So I had to use PowerMock for that. Anyway, could you please take a look at my PR https://github.com/folio-org/mod-feesfines/pull/88? Especially at classes `TenantRefAPI`, `PubSubService` and `APITests`. If you find them adequate, maybe you could reuse some of that work for your implementation? You can completely disregard all test classes except `AccountsAPITest` and `TenantRefAPITest`, as they only contain changes related to refactoring.

marcjohnson-kint:
@bohdan-suprun  @OleksandrVidinieiev Thank you for working together to consolidate a common base line for pub-sub integration that both changes can use

bohdan-suprun:
@OleksandrVidinieiev, thanks, I've reused a lot of your code. Please take a look at this [line ](https://github.com/folio-org/mod-feesfines/pull/89/files#diff-4dc6189cf737611cc3b5106dc4c55148R45) and  at the `AccountAPITest` (just in case I missed something).

src/main/java/org/folio/rest/client/FeeFinePubSubClient.java
```diff
@@ -0,0 +1,71 @@
...
+import org.folio.rest.exception.InternalServerError;
+import org.folio.rest.jaxrs.model.Event;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.vertx.core.Vertx;
+import io.vertx.core.buffer.Buffer;
+import io.vertx.ext.web.client.HttpResponse;
+
+public class FeeFinePubSubClient {
```
OleksandrVidinieiev, line 18: Looks like this class and `PubSubService` are doing the same thing. Should we merge them together?

src/main/resources/MessagingDescriptor.json
```diff
@@ -5,8 +5,14 @@
       "description": "Patron fee/fine balance was changed",
       "eventTTL": 1,
       "signed": false
+    },
+    {
+      "eventType": "FF_ACCOUNT_WITH_LOAN_CLOSED",
+      "description": "An account with a loan associated is closed",
+      "eventTTL": 30,
```
OleksandrVidinieiev, line 12: Currently `eventTTL` for all events is hardcoded to `1`, see `PubSubService#createEvent()`. What does it do anyway? There is no mention of it in mod-pubsub documentation...

bohdan-suprun: As far as I know from the Folijet team this flag is not implemented yet. It will control how long the message will 'live' in the queue. 

I'll probably revert it to 1 as well.

src/main/java/org/folio/rest/client/FeeFinePubSubClient.java
bohdan-suprun: For the `account with loan closed event` we need to forward failures to the end-user in case event publish was rejected.

However, pub/sub rejects event publish with code 400 in case there are no subscribers for this event. By introducing this client, we're overriding this behavior, so it is treated as success. In few words, it is  an equivalent of the `PubSubClientUtils.sendEventMessage` method from the `pub-sub-client` lib. I'd keep it as separate class.

What do you think?

src/main/resources/MessagingDescriptor.json
```diff
@@ -5,8 +5,14 @@
       "description": "Patron fee/fine balance was changed",
       "eventTTL": 1,
       "signed": false
+    },
+    {
+      "eventType": "FF_ACCOUNT_WITH_LOAN_CLOSED",
```
OleksandrVidinieiev, line 10: We are already issueing an event `FF_BALANCE_CHANGED` when an account is closed. Can we reuse that for when an account related to a loan is closed? We can add `loanId` to the existing event's payload.

bohdan-suprun: Hmm... That's a good point.

The idea in publishing this event is to be able close the associated loan when all `lost item fees` are closed/resolved.

So, mod-circulation will subscribe to this event type. I'm afraid that this will overuse circulation computing resources, because the `FF_BALANCE_CHANGED` event will be generated when an account (either automated|manual) created/updated (when paid/waived/transferred partially|fully)/deleted. And for each generated event we will have to:
# Fetch the loan;
# Fetch all associated accounts.

By introducing the `FF_ACCOUNT_WITH_LOAN_CLOSED` event, we minimize amount of events, so we'll receive most likely events that we're interesting in  (closed and automated).

Does it make sense?

src/main/resources/MessagingDescriptor.json
OleksandrVidinieiev: Do you happen to also know which time unit it is supposed to represent? (seconds, minutes, etc.)

bohdan-suprun: it is time to live in minutes.

src/main/resources/MessagingDescriptor.json
OleksandrVidinieiev: You are right, events for closed "acccounts" related to loans are just a small subset of events of type `FF_BALANCE_CHANGE` issued by mod-feesfines.

OleksandrVidinieiev: Is it too late to change the name of the event to `FEEFINE_WITH_LOAN_CLOSED`? A major refactoring of mod-feesfines is planned, which will involve revision of terminology used by the module. So the term `account` is likely to be replaced with something that makes a bit more sense.

marcjohnson-kint: > Is it too late to change the name of the event to `FEEFINE_WITH_LOAN_CLOSED`? 

What would you suggest as an alternative?

> A major refactoring of mod-feesfines is planned, which will involve revision of terminology used by the module. So the term `account` is likely to be replaced with something that makes a bit more sense.

I'm glad that this is being done. Please can you refer me to further information about this e.g. a JIRA issue, planned domain model etc.

bohdan-suprun: @OleksandrVidinieiev, `FEEFINE_WITH_LOAN_CLOSED` is fine for me, will replace. Thanks.

OleksandrVidinieiev: > What would you suggest as an alternative?
This is the alternative. The original name was `FF_ACCOUNT_WITH_LOAN_CLOSED`.

marcjohnson-kint: @OleksandrVidinieiev Does that mean that the new terminology is going to be `fee / fine` rather than `account`?

src/main/java/org/folio/rest/client/FeeFinePubSubClient.java
OleksandrVidinieiev: Thanks for pointing this 400 error out, I did not consider this case. Handling the error for missing subscribers correctly is a good idea. But you also made me think if we should fail the whole operation if event publishing failed in other cases too...

I will resolve this comment.

src/main/resources/MessagingDescriptor.json
```diff
@@ -5,8 +5,14 @@
       "description": "Patron fee/fine balance was changed",
       "eventTTL": 1,
       "signed": false
+    },
+    {
+      "eventType": "FEEFINE_WITH_LOAN_CLOSED",
```
marcjohnson-kint, line 10: Does this event refer to a fee / fine that has a loan being closed, or something to do with a fee / fine with a closed loan?

bohdan-suprun: this event refers to a fee/fine that was recently closed and has a loan associated.

marcjohnson-kint: > this event refers to a fee/fine that was recently closed and has a loan associated.

Why is it important to have separate events for fees / fines that are closed with a link to a loan and fees / fines that are closed without a link to a loan?

Can it be generalised to `fee / fine closed` and optionally including the linked loan ID in the event representation?

bohdan-suprun: This will increase load on mod-circulation. Each time any account is closed pub/sub will notify mod-circulation about this. 

An account _without_ a loan it is a manual charge and mod-circulation is not interesting in such events. 

Should I proceed anyway?

src/main/resources/MessagingDescriptor.json
OleksandrVidinieiev: @marcjohnson-kint The new terminology does not exist yet. But if we change the term `account` to something else, event `FEEFINE_ACCOUNT_WITH_LOAN_CLOSED` will definitely need to be renamed too. On the other hand, stripping this name of the `ACCOUNT` part now will:
a) preserve its meaning: account is an instance of a fee/fine, closing a fee/fine means closing an account
b) eliminate the need to rename the event after `account` is renamed to something else

Does it make sense?

src/main/java/org/folio/rest/service/AccountService.java
```diff
@@ -0,0 +1,74 @@
...
+        return pubSubService.publishFeeFineWithLoanClosedEvent(account)
+          .thenApply(notUsed -> responseResult);
+      }
+
+      return completedFuture(responseResult);
+    }).exceptionally(error -> {
+      log.error("Cannot publish fee/fine closed event [loanId - {}, feeFineId - {}]," +
+        " error occurred {}", account.getLoanId(), account.getId(), error);
+
+      return succeededFuture(respond500WithTextPlain(error.getMessage()));
```
OleksandrVidinieiev, line 57: Correct me if I'm wrong, but if the update of an account in the DB was successful, but the event publishing failed, we would still return 500, right? Is this the desired behavior?

src/main/resources/MessagingDescriptor.json
marcjohnson-kint: > This will increase load on mod-circulation. Each time any account is closed pub/sub will notify mod-circulation about this.
> 

That's true, there could be additional load. Can circulation complete the request quickly if there is a no loan ID?

> An account _without_ a loan it is a manual charge and mod-circulation is not interesting in such events.

Are some manual charges related to loans e.g. actual cost fees?

> Should I proceed anyway?

It might be that we want a specialised event. I'd prefer to wait to decide that, unless we can think of a really explicit name like `LOAN_RELATED_FEE_FINE_CLOSED`?

src/main/java/org/folio/rest/service/AccountService.java
marcjohnson-kint: FOLIO does not yet have guidance on what should happen when a process can complete except for the publishing of events. There is a group working on this.

In the mean time, I think we agreed to fail so that the user would at least be aware that any follow on process will not happen.

src/main/resources/MessagingDescriptor.json
marcjohnson-kint: @OleksandrVidinieiev Thanks for explaining that more.

I'm ok with removing the account aspect now, and then reviewing if the names need to change further when the domain analysis has been done.

(The other day the term `bill` was mentioned to me for referring to what is issued when a fee or fine is charged to a patron)

src/main/resources/MessagingDescriptor.json
bohdan-suprun: > Can circulation complete the request quickly if there is a no loan ID?

Sure, it can.

src/main/resources/MessagingDescriptor.json
OleksandrVidinieiev: @marcjohnson-kint I've heard this entity being referred to as _bill_ too. If that's an established term, it makes it a likely candidate for _account_ replacement.

src/main/resources/MessagingDescriptor.json
bohdan-suprun: > Are some manual charges related to loans e.g. actual cost fees?

Yes, actual cost fees are going to be manual charges, but I'd be good to have a link to the loan for them as well.

src/main/java/org/folio/rest/service/AccountService.java
OleksandrVidinieiev: What is this group? Is there a way for us to track progress of their work?

marcjohnson-kint: The group is made up of Taras, Craig, Vince, Jakub and myself at the moment. There is the outline of a working document (I've haven't had chance to update it for a week or so) - https://wiki.folio.org/display/DD/Handling+Failure+in+Asynchronous+Processes

It is likely this group will report progress to the Technical Leads and the Technical Council as well.

src/main/resources/MessagingDescriptor.json
marcjohnson-kint: > This may result into a situation when the event payload will be constantly updated with some extra-fields, that needed for a particular subscriber, which may lead to another one dependency on the payload version.
> 
> I'd prefer to have specialized event rather then an generalized event with extra-properties in payload.

Ok, I'm happy to go with that, as long as we can make the event name very specific and unambiguous.

src/main/resources/MessagingDescriptor.json
marcjohnson-kint: > @marcjohnson-kint I've heard this entity being referred to as _bill_ too. If that's an established term, it makes it a likely candidate for _account_ replacement.

I don't know if it is an established term. I heard it for the first time last week. I'd like it to be considered during the analysis of the domain.

src/main/java/org/folio/rest/service/AccountService.java
OleksandrVidinieiev: @marcjohnson-kint thank you

src/main/resources/MessagingDescriptor.json
bohdan-suprun: Ok, I'll proceed with `LOAN_RELATED_FEE_FINE_CLOSED` name then?

marcjohnson-kint: > Ok, I'll proceed with `LOAN_RELATED_FEE_FINE_CLOSED` name then?

I'm ok with that. I'm happy for you to think of a better name. The most important aspect is that it is unambiguous about what is closed and how the loan is involved.

src/main/java/org/folio/rest/client/FeeFinePubSubClient.java
```diff
@@ -0,0 +1,71 @@
...
+  private static final Logger log = LoggerFactory.getLogger(FeeFinePubSubClient.class);
+  private final OkapiClient okapiClient;
+
+  public FeeFinePubSubClient(Vertx vertx, Map<String, String> okapiHeaders) {
```
marcjohnson-kint, line 22: Does the prebuilt pub-sub client not do this already, or is it inappropriate because it does not give enough control over interpreting the response?

bohdan-suprun: The prebuilt pub/sub client does not ignore the 400 `No subscribers for the event type` error. But this overridden client ignores it.

marcjohnson-kint: Thanks @bohdan-suprun

bohdan-suprun:
Hi @marcjohnson-kint, do you have any comments for this PR? Should I proceed with merge?

marcjohnson-kint:
> Hi @marcjohnson-kint, do you have any comments for this PR? Should I proceed with merge?

I've been trying to review it for the last few days. I will try to review it today. Apologies for the delay :-/

src/main/java/org/folio/rest/service/PubSubService.java
```diff
@@ -62,6 +67,11 @@ public PubSubService(Map<String, String> okapiHeaders, Vertx vertx) {
     return publishAccountBalanceChangeEvent(account);
   }
 
+  public CompletableFuture<Void> publishLoanRelatedFeeFineClosedEvent(Account feeFine) {
+    return pubSubClient.publishEvent(createEvent(LOAN_RELATED_FEE_FINE_CLOSED,
+      LoanRelatedFeeFineClosedEvent.forFeeFine(feeFine).toJsonString()));
+  }
+
   private CompletableFuture<Boolean> publishEvent(final Event event) {
     return PubSubClientUtils.sendEventMessage(event, connectionParams)
```
marcjohnson-kint, line 76: What is the different between the `FeeFinePubSubClient` and the `PubSubClientUtils`?

src/main/resources/MessagingDescriptor.json
```diff
@@ -5,8 +5,14 @@
       "description": "Patron fee/fine balance was changed",
       "eventTTL": 1,
       "signed": false
+    },
+    {
+      "eventType": "LOAN_RELATED_FEE_FINE_CLOSED",
+      "description": "An account with a loan associated is closed",
```
marcjohnson-kint, line 11: Given we are avoiding the term `account` in the event name, I think we could also avoid it in the description.

I think `is closed` is a statement about the current state, whereas `has been` might be more appropriate for an event, given they represent facts about what has happened in the past.

Does that make sense? What do you think?

```suggestion
      "description": "A fee or fine with a loan associated has been closed",
```

src/test/java/org/folio/test/support/matcher/MappableMatcher.java
```diff
@@ -0,0 +1,38 @@
+package org.folio.test.support.matcher;
+
+import java.util.function.Function;
+
+import org.hamcrest.Description;
+import org.hamcrest.Matcher;
+import org.hamcrest.TypeSafeDiagnosingMatcher;
+
+public final class MappableMatcher<T, V> extends TypeSafeDiagnosingMatcher<T> {
```
marcjohnson-kint, line 9: What is a mappable matcher? What does mappable mean in this context?

src/test/java/org/folio/test/support/matcher/AccountMatchers.java
```diff
@@ -0,0 +1,24 @@
...
+
+import org.hamcrest.Matcher;
+
+import io.restassured.response.Response;
+
+public final class AccountMatchers {
+
+  private AccountMatchers() {}
+
+  public static Matcher<Response> isAccountPaidFully() {
```
marcjohnson-kint, line 15: I imagine that `account` is implicit in the context of the use of this matcher, and so might it read better without that in the matchers name e.g. `account is paid fully` rather than `account is account paid fully`

What do you think?

```suggestion
  public static Matcher<Response> isPaidFully() {
```

src/test/java/org/folio/rest/utils/TestResourceClient.java
```diff
@@ -0,0 +1,65 @@
+package org.folio.rest.utils;
+
+import static org.folio.rest.impl.APITests.getOkapiUrl;
+
+import org.folio.HttpStatus;
+
+import io.restassured.response.Response;
+
+public class TestResourceClient {
```
marcjohnson-kint, line 9: Could this be called `ResourceClient` instead? Is it important that this is only for tests, given that it is defined within the test code source set?

src/main/java/org/folio/rest/domain/LoanRelatedFeeFineClosedEvent.java
```diff
@@ -0,0 +1,31 @@
+package org.folio.rest.domain;
+
+import org.folio.rest.jaxrs.model.Account;
+
+import io.vertx.core.json.JsonObject;
+
+public final class LoanRelatedFeeFineClosedEvent {
```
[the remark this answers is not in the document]
bohdan-suprun: An `Event` is a pub/sub entity which contains some configuration properties as well as payload that attached to the event. 

This class is the payload of the event.

src/main/java/org/folio/rest/service/PubSubService.java
bohdan-suprun: `PubSubClientUtils` is a class from `mod-pubsub-utils` library, which fails when there are no subscribers for the event. 

`FeeFinePubSubClient` - overrides this behavior and treats this case as a success.

src/test/java/org/folio/test/support/matcher/MappableMatcher.java
bohdan-suprun: This matcher is used to map something to a value and then assert this mapped value. 

For example:

Jayway `hasJsonPath` matchers can assert json as string only, however we usually use `JsonObject`. We can now use following logic in order to assert an JsonObject with string `hasJsonPath` matcher:

```
assertThat(jsonObject, new MappableMatcher<JsonObject, String>(JsonObject::toString,
allOf(
   hasJsonPath(...),
   hasJsonPath(...)
)));
```

src/main/java/org/folio/rest/domain/FeeFineAmount.java
```diff
@@ -0,0 +1,21 @@
...
+      ? BigDecimal.valueOf(amount)
+      : BigDecimal.ZERO;
+  }
+
+  public final boolean hasZeroAmount() {
+    return getScaledAmount().compareTo(BigDecimal.ZERO) == 0;
+  }
+
+  private BigDecimal getScaledAmount() {
+    return amount.setScale(2, BigDecimal.ROUND_HALF_UP);
```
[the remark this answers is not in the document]
bohdan-suprun: I've added new paragraph `Implementation details | Fee/fine amount rounding`

src/main/java/org/folio/rest/service/AccountService.java
```diff
@@ -0,0 +1,74 @@
...
+import org.folio.rest.domain.FeeFineStatus;
+import org.folio.rest.domain.FeeFineAmount;
+import org.folio.rest.jaxrs.model.Account;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.vertx.core.AsyncResult;
+import io.vertx.core.Context;
+
+public class AccountService {
```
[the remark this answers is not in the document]
bohdan-suprun: Renamed to `AccountUpdateService`

src/main/java/org/folio/rest/service/AccountService.java
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  private boolean isFeeFineWithLoanClosed(Account feeFine) {
+    final FeeFineAmount feeFineAmount = new FeeFineAmount(feeFine.getRemaining());
+    final FeeFineStatus feeFineStatus = feeFine.getStatus() != null
+      ? forValue(feeFine.getStatus().getName()) : null;
+
+    return feeFineStatus == CLOSED && StringUtils.isNotBlank(feeFine.getLoanId())
+      && feeFineAmount.hasZeroAmount();
+  }
```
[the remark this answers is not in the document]
bohdan-suprun: Split to `isFeeFineClosed` and `isFeeFineAssociatedToLoan`

src/main/resources/MessagingDescriptor.json
bohdan-suprun: Agree, updated.

src/test/java/org/folio/test/support/matcher/AccountMatchers.java
bohdan-suprun: renamed.

src/test/java/org/folio/rest/utils/TestResourceClient.java
bohdan-suprun: It is renamed now. The `Test` part was just to not mix up production client with the test client.

src/main/java/org/folio/rest/service/PubSubService.java
```diff
@@ -25,13 +28,15 @@
 public class PubSubService {
```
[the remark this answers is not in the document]
bohdan-suprun: Sure, split to 3 classes:
* EventPublisher - low level publishing methods (communicates with pub/sub).
* AccountEventPublisher - all account-related events;
* PubSubRegistrationService - module registration in pub/sub

src/main/java/org/folio/rest/domain/LoanRelatedFeeFineClosedEvent.java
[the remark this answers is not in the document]
marcjohnson-kint: Ah, ok. This might be where the naming is a little confusing. The pub-sub `Event` is really an `EventMessage` or an `EventNotification` rather than the event itself.

As that cannot be changed because it is part of a the library (I assume) I'll resolve this as out of scope.

src/main/java/org/folio/rest/service/AccountService.java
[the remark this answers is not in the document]
marcjohnson-kint: Thanks

src/test/java/org/folio/test/support/matcher/MappableMatcher.java
marcjohnson-kint: This converts the `JsonObject` to a `String` and then delegates the rest of the matching to other matchers?

Ok, maybe this could be renamed `TypeMappingMatcher` or it could be made to be more specific to the JSON Path use case and renamed JSONPathMatcher?

src/main/java/org/folio/rest/impl/TenantRefAPI.java
```diff
@@ -52,7 +52,7 @@ public void postTenant(TenantAttributes tenantAttributes,
           }
 
           vertx.executeBlocking(
-            promise -> new PubSubService(headers, vertx).registerModuleInPubsub(promise),
+            promise -> new PubSubRegistrationService(vertx, headers).registerModuleInPubsub(promise),
```
marcjohnson-kint, line 55: Given that the class name now refers to pub-sub registration, the method could be shortened to `registerModule`

```suggestion
            promise -> new PubSubRegistrationService(vertx, headers).registerModule(promise),
```

src/main/java/org/folio/rest/service/AccountUpdateService.java
```diff
@@ -0,0 +1,81 @@
...
+    return feeFineStatus == CLOSED && feeFineAmount.hasZeroAmount();
+  }
+
+  private boolean isFeeFineAssociatedToLoan(Account feeFine) {
+    return StringUtils.isNotBlank(feeFine.getLoanId());
+  }
+
+  private boolean isFeeFineUpdateSucceeded(AsyncResult<Response> responseAsyncResult) {
+    return responseAsyncResult.succeeded()
+      || responseAsyncResult.result().getStatus() == HTTP_NO_CONTENT.toInt();
```
marcjohnson-kint, line 79: Should this be and (`&&`) rather than  or (`||`)? 

The update succeeded if either the result succeeded or the HTTP status was No Content?

src/test/java/org/folio/rest/impl/APITests.java
```diff
@@ -1,99 +1,178 @@
...
+      .willReturn(created()));
+    wireMock.stubFor(post(urlEqualTo("/pubsub/event-types"))
+      .atPriority(100)
+      .willReturn(created()));
+
+    // forward everything to okapi
+    wireMock.stubFor(any(anyUrl())
+      .atPriority(Integer.MAX_VALUE)
+      .willReturn(aResponse().proxiedFrom("http://localhost:" + OKAPI_PORT)));
+  }
```
marcjohnson-kint, line 119: Could this be extracted to a class of it's own?

src/main/java/org/folio/rest/service/AccountUpdateService.java
bohdan-suprun: Yes, thank you

src/test/java/org/folio/rest/impl/APITests.java
bohdan-suprun: Moved to OkapiDeployment class.

src/test/java/org/folio/test/support/matcher/MappableMatcher.java
bohdan-suprun: Renamed to TypeMappingMatcher

bohdan-suprun:
> I think the AccountsApiTests could do with quite a bit of refinement / clean up, however I think that is too much work for this pull request.

I'm not sure if I understand what should be done here.

PR #101, opened 2020-07-23
[MODFEE-87] Add currency formatter

OleksandrVidinieiev:
## Approach 
- fix issue with decimal formatting for double values;
- update tests;

Resolves: [MODFEE-87](https://issues.folio.org/browse/MODFEE-87)


src/test/java/org/folio/rest/impl/FeeFineActionsAPITest.java
```diff
@@ -81,6 +83,9 @@
   private static final String PRIMARY_CONTRIBUTOR_NAME = "Primary contributor";
   private static final String NON_PRIMARY_CONTRIBUTOR_NAME = "Non-primary contributor";
 
+  private static final NumberFormat CURRENCY_FORMATTER = new DecimalFormat("#0.00");
+
+
```
roman-barannyk, line 88: extra line?

OleksandrVidinieiev: removed

PR #111, opened 2020-08-26
[MODFEE-84] Create an endpoint for waiving fees/fines

OleksandrVidinieiev:
Resolves [MODFEE-84](https://issues.folio.org/browse/MODFEE-84).


src/main/java/org/folio/rest/impl/AccountsAPI.java
```diff
@@ -425,31 +416,58 @@ public void postAccountsPayByAccountId(String accountId, ActionRequest request,
 
     new DefaultActionService(okapiHeaders, vertxContext)
       .pay(accountId, request)
-      .onSuccess(context -> {
-        ActionSuccessResponse response = new ActionSuccessResponse()
+      .onComplete(result -> handleActionResult(accountId, request, result, asyncResultHandler,
+        PostAccountsPayByAccountIdResponse::respond201WithApplicationJson,
+        PostAccountsPayByAccountIdResponse::respond422WithApplicationJson,
+        PostAccountsPayByAccountIdResponse::respond404WithTextPlain,
+        PostAccountsPayByAccountIdResponse::respond500WithTextPlain));
```
alexanderkurash, line 423: Extract to a class

src/test/java/org/folio/rest/utils/ResourceClients.java
```diff
@@ -9,23 +9,31 @@ public static ResourceClient accountsClient() {
...
   }
 
   public static ResourceClient accountsCheckRefundClient(String accountId) {
-    return new ResourceClient("/accounts/" + accountId + "/check-refund");
+    return getUrlForAction(accountId, "check-refund");
   }
 
   public static ResourceClient accountsPayClient(String accountId) {
-    return new ResourceClient("/accounts/" + accountId + "/pay");
+    return getUrlForAction(accountId, "pay");
```
alexanderkurash, line 28: make? create? build?

src/main/java/org/folio/rest/impl/AccountsAPI.java
OleksandrVidinieiev: extracted

src/test/java/org/folio/rest/utils/ResourceClients.java
OleksandrVidinieiev: renamed

alexanderkurash (approved):

PR #115, opened 2020-09-08
[MODFEE-86] Create refund endpoint

OleksandrVidinieiev:
Resolves [MODFEE-86](https://issues.folio.org/browse/MODFEE-86).

### DIFF
--- a/src/main/java/org/folio/rest/repository/FeeFineActionRepository.java
+++ b/src/main/java/org/folio/rest/repository/FeeFineActionRepository.java
@@ -1,14 +1,22 @@
 package org.folio.rest.repository;
 
+import static java.util.stream.Collectors.toList;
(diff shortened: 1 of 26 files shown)


src/main/java/org/folio/rest/repository/FeeFineActionRepository.java
```diff
@@ -37,6 +49,14 @@ public FeeFineActionRepository(PostgresClient pgClient) {
       .map(Results::getResults);
   }
 
+  public Future<List<Feefineaction>> findRefundableActionsForAccount(String accountId) {
+    return findActionsForAccount(accountId)
```
alexanderkurash, line 53: Limit of 10?

alexanderkurash (approved):

PR #116, opened 2020-09-23
[MODFEE-86] Fix check-refund logic

OleksandrVidinieiev:
Resolves [MODFEE-86](https://issues.folio.org/browse/MODFEE-86).

### DIFF
--- a/src/main/java/org/folio/rest/service/action/validation/ActionValidationResult.java
+++ b/src/main/java/org/folio/rest/service/action/validation/ActionValidationResult.java
@@ -12,8 +12,7 @@ public class ActionValidationResult {
   }
 
   public ActionValidationResult(MonetaryValue remainingAmount, MonetaryValue requestedAmount) {
-    this.remainingAmount = remainingAmount.toString();
-    this.requestedAmount = requestedAmount.toString();
+    this(remainingAmount.toString(), requestedAmount.toString());
   }
 
   public String getRemainingAmount() {
--- a/src/main/java/org/folio/rest/service/action/validation/ActionValidationResult.java
+++ b/src/main/java/org/folio/rest/service/action/validation/ActionValidationResult.java
@@ -4,23 +4,23 @@ import org.folio.rest.domain.MonetaryValue;
 
 public class ActionValidationResult {
   private final String remainingAmount;
-  private final String formattedAmount;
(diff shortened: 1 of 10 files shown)


src/main/java/org/folio/rest/service/action/validation/ActionValidationResult.java
```diff
@@ -9,6 +11,11 @@ public ActionValidationResult(String remainingAmount, String formattedAmount) {
     this.formattedAmount = formattedAmount;
   }
 
+  public ActionValidationResult(MonetaryValue remainingAmount, MonetaryValue formattedAmount) {
+    this.remainingAmount = remainingAmount.toString();
+    this.formattedAmount = formattedAmount.toString();
```
alexanderkurash, line 16: I'm not sure about the name, makes it look like this variable is about formatting. Will `actionableAmount` work (as a generic version of `refundableAmount`)?

src/main/java/org/folio/rest/service/action/validation/DefaultActionValidationService.java
```diff
@@ -40,8 +40,10 @@ protected void validateAccountStatus(Account account) {
   }
 
   @Override
-  protected MonetaryValue calculateRemainingBalance(Account account, MonetaryValue requestedAmount) {
-    return new MonetaryValue(account.getRemaining()).subtract(requestedAmount);
+  protected Future<MonetaryValue> calculateRemainingBalance(Account account,
```
alexanderkurash, line 43: Do we really need to "futurize" this method?

OleksandrVidinieiev: Yes, calculation of `remainingAmount` is based on calculation of refundable amount, which requires calling the DB, thus making the whole process asynchronous.

src/main/java/org/folio/rest/service/action/validation/ActionValidationResult.java
alexanderkurash: I think `requestedAmount` is fine, it looks like we're using it when this object is created - `new ActionValidationResult(remainingBalance, requestedAmount)` so it would make sense if these names matched

alexanderkurash (approved):

PR #127, opened 2020-10-08
Create cancel endpoint for multiple selected accounts

mrdpopov:
## Purpose

Create cancel endpoint for multiple selected accounts

### DIFF
--- a/ramls/actions/cancelBulkActionRequest.json
+++ b/ramls/actions/cancelBulkActionRequest.json
@@ -33,9 +33,9 @@
   "additionalProperties": false,
   "required": [
     "accountIds",
-    "amount",
     "notifyPatron",
     "servicePointId",
-    "userName"
+    "userName",
+    "comments"
   ]
 }
--- a/ramls/actions/cancelBulkActionRequest.json
+++ b/ramls/actions/cancelBulkActionRequest.json
@@ -1,0 +1,40 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "description": "Bulk cancel action request",
+  "type": "object",
+  "properties": {
+    "accountIds": {
+      "type": "array",
+      "items": {
+        "type": "string",
+        "$ref": "../uuid.json"
+      },
+      "description": "IDs of multiple accounts"
+    },
+    "comments": {
+      "type": "string",
+      "description": "Additional info for staff or patron. Format: [STAFF : staff comment \n PATRON : patron comment]"
+    },
+    "notifyPatron": {
+      "type": "boolean",
+      "description": "Should send patron notification or not"
+    },
+    "servicePointId": {
+      "type": "string",
+      "$ref": "../uuid.json",
+      "description": "Unique ID of the service point where the payment was made"
+    },
+    "userName": {
+      "type": "string",
+      "description": "Name of the user that was logged in when the action was performed"
+    }
+  },
+  "additionalProperties": false,
+  "required": [
+    "accountIds",
+    "amount",
+    "notifyPatron",
+    "servicePointId",
+    "userName"
+  ]
+}
--- a/ramls/actions/cancelBulkActionRequest.json
+++ b/ramls/actions/cancelBulkActionRequest.json
@@ -2,6 +2,7 @@
   "$schema": "http://json-schema.org/draft-04/schema#",
   "description": "Bulk cancel action request",
   "type": "object",
+  "javaInterfaces": ["org.folio.rest.domain.BulkActionRequest"],
   "properties": {
     "accountIds": {
       "type": "array",
--- a/src/main/java/org/folio/rest/service/action/BulkActionService.java
+++ b/src/main/java/org/folio/rest/service/action/BulkActionService.java
@@ -57,6 +57,22 @@ public abstract class BulkActionService {
     this.amountSplitterStrategy = new SplitEvenlyRecursively();
   }
 
+  public BulkActionService(Action action, ActionValidationService validationService,
+                           BulkActionAmountSplitterStrategy bulkActionAmountSplitterStrategy,
+                           Map<String, String> headers, Context context) {
+
+    PostgresClient postgresClient = getInstance(context.owner(), tenantId(headers));
+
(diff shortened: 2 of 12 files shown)


ramls/actions/cancelBulkActionRequest.json
```diff
@@ -0,0 +1,41 @@
...
+    },
+    "userName": {
+      "type": "string",
+      "description": "Name of the user that was logged in when the action was performed"
+    }
+  },
+  "additionalProperties": false,
+  "required": [
+    "accountIds",
+    "amount",
```
alexanderkurash, line 36: There's no `amount` field in this schema. Single-account cancel requires these fields:
    "comments",
    "notifyPatron",
    "servicePointId",
    "userName"

so I think it should be the same list here, with an addition of `accountIds`

src/main/java/org/folio/rest/service/action/BulkActionService.java
```diff
@@ -57,6 +57,22 @@ public BulkActionService(Action action, ActionValidationService validationServic
     this.amountSplitterStrategy = new SplitEvenlyRecursively();
   }
 
+  public BulkActionService(Action action, ActionValidationService validationService,
+                           BulkActionAmountSplitterStrategy bulkActionAmountSplitterStrategy,
+                           Map<String, String> headers, Context context) {
```
alexanderkurash, line 62: Parameters indentation

src/main/java/org/folio/rest/service/action/BulkActionService.java
```diff
@@ -67,6 +83,7 @@ public BulkActionService(Action action, ActionValidationService validationServic
   }
 
   private Future<BulkActionContext> findAccounts(BulkActionContext context) {
+
```
alexanderkurash, line 86: Why do we need this blank line?

src/main/java/org/folio/rest/service/action/BulkActionService.java
```diff
@@ -79,7 +96,8 @@ public BulkActionService(Action action, ActionValidationService validationServic
   }
 
   protected Future<BulkActionContext> createFeeFineActions(BulkActionContext context) {
-    final DefaultBulkActionRequest request = (DefaultBulkActionRequest) context.getRequest();
+
```
alexanderkurash, line 99: Same question here

src/main/java/org/folio/rest/service/action/BulkActionService.java
```diff
@@ -110,18 +127,22 @@ protected Feefineaction createFeeFineActionAndUpdateAccount(Account account, Mon
...
       .withBalance(remainingAmountAfterAction.toDouble())
       .withTypeAction(actionType)
       .withId(UUID.randomUUID().toString())
       .withDateAction(new Date());
 
+    if (request instanceof DefaultBulkActionRequest) {
+      DefaultBulkActionRequest defaultBulkActionRequest = (DefaultBulkActionRequest) request;
+      feeFineAction.setPaymentMethod(defaultBulkActionRequest.getPaymentMethod());
+      feeFineAction.setNotify(defaultBulkActionRequest.getNotifyPatron());
+      feeFineAction.setTransactionInformation(defaultBulkActionRequest.getTransactionInfo());
```
alexanderkurash, line 143: We can also do it with a `.with()` chain to keep the same style as above.

src/main/java/org/folio/rest/service/action/BulkCancelActionService.java
```diff
@@ -0,0 +1,36 @@
...
+public class BulkCancelActionService extends BulkActionService {
+
+  public BulkCancelActionService(Map<String, String> headers, Context context) {
+    super(Action.CANCEL, new CancelActionValidationService(headers, context),
+      new DummyBulkSplitterStrategy(), headers, context);
+  }
+
+  @Override
+  protected Future<BulkActionContext> validateAction(BulkActionContext context) {
+
```
alexanderkurash, line 31: Blank line..

ramls/actions/cancelBulkActionRequest.json
mrdpopov: done.

src/main/java/org/folio/rest/service/action/BulkActionService.java
mrdpopov: done.

src/main/java/org/folio/rest/service/action/BulkActionService.java
mrdpopov: done.

src/main/java/org/folio/rest/service/action/BulkActionService.java
mrdpopov: No we can not. There will be runtime errors in case of bulk cancel request, which is not instance of DefaultBulkActionRequest

src/main/java/org/folio/rest/service/action/BulkCancelActionService.java
mrdpopov: done.

src/main/java/org/folio/rest/service/action/BulkActionService.java
mrdpopov: done.

src/main/java/org/folio/rest/service/action/BulkActionService.java
alexanderkurash: I meant we can put a new `.with()` chain inside this `if` instead of `set`s, just for these three fields. It just looks a little bit cleaner and uniform with what's above.

mrdpopov: I don't think it will be better because it will look like this: 
`feeFineAction = feeFineAction.withPaymentMethod(defaultBulkActionRequest.getPaymentMethod());`

besides we'll have to remove final modifier.

mrdpopov: I'll override this method.

alexanderkurash (approved):

PR #136, opened 2020-10-21
MODFEE-131 - Fee/fine actions do not appear in circ log

siarhei-charniak:
[MODFEE-131](https://issues.folio.org/browse/MODFEE-131) - Fee/fine actions do not appear in circ log


src/main/java/org/folio/rest/service/LogContextService.java
```diff
@@ -1,87 +1,40 @@
...
-    return succeededFuture(context);
+      .compose(account -> createFeeFineLogContext(action, account));
   }
 
-  private Future<FeeFineLogContext> addFeeFineData(Feefine feefine) {
-    ofNullable(feefine).ifPresent(f -> context = context.withAutomated(f.getAutomatic()));
-    return succeededFuture(context);
+  public Future<JsonObject> createFeeFineLogContext(Feefineaction action, Account account) {
+    return feeFineRepository.getById(account.getFeeFineId())
+      .compose(feeFine -> usersClient.fetchUserById(account.getUserId())
```
khandramai, line 37: What is a reason to fetch user by id here and introduce additional call here? As I remember we discussed fetching user in the mod-audit to don't affect other business flows.

src/main/java/org/folio/rest/impl/FeeFineActionsAPI.java
```diff
@@ -147,12 +149,16 @@ private Response sendPatronNoticeIfNeedBe(Feefineaction entity, Map<String, Stri
...
-      .map(context -> {
-        new LogEventPublisher(vertxContext, okapiHeaders).publishLogEvent(context.asJson(), FEE_FINE);
-        return Future.succeededFuture();
-      })
-      .map(v -> response);
+    // do not publish log records for CREDIT and REFUND actions
+    if (!CREDIT.isActionForResult(entity.getTypeAction()) && !REFUND.isActionForResult(entity.getTypeAction())) {
+      return new LogContextService(vertxContext.owner(), okapiHeaders).createFeeFineLogContext(entity)
+        .compose(context -> {
+          new LogEventPublisher(vertxContext, okapiHeaders).publishLogEvent(context, FEE_FINE);
```
khandramai, line 156: Maybe it is more suitable to use name event instead of context?

src/main/java/org/folio/rest/service/action/ActionService.java
```diff
@@ -168,4 +179,16 @@ protected Feefineaction createFeeFineActionAndUpdateAccount(Account account, Mon
     return succeededFuture(context);
   }
 
+  private Future<ActionContext> publishLogEvents(ActionContext actionContext) {
+    return all(actionContext.getFeeFineActions().stream()
+      // do not publish log records for CREDIT and REFUND actions
+      .filter(ffa -> !CREDIT.isActionForResult(ffa.getTypeAction()) && !REFUND.isActionForResult(ffa.getTypeAction()))
+      .map(ffa -> logEventService.createFeeFineLogEventPayload(ffa, actionContext.getAccounts().get(ffa.getAccountId()))
+        .compose(eventPayload -> {
+          logEventPublisher.publishLogEvent(eventPayload, FEE_FINE);
```
khandramai, line 188: Is this published async? Maybe it makes sense to make CompletableFuture.runAsync() and verify somehow by introduction of Thread.sleep()?

khandramai (approved):

PR #137, opened 2020-10-23
MODFEE-114: Add Lost item fee (actual cost) type.

bohdan-suprun:
https://issues.folio.org/browse/MODFEE-114 - BACKEND - Add ACTUAL COST automated Fee/fine type to Manual charges table.

### Changes:

* Add ` Lost item fee (actual cost)` fee/fine type.
* Add some tests to verify the change.


src/main/resources/templates/db_scripts/schema.json
```diff
@@ -9,7 +9,7 @@
     },
     {
       "tableName":"feefines",
-      "fromModuleVersion":"15.7.1",
+      "fromModuleVersion":"15.10.0",
       "withMetadata":true,
       "customSnippetPath": "populate-feefines.sql",
```
marcjohnson-kint, line 14: I didn't think we changed the fromModuleVersion of a table for a configuration change. 

Is this so that the snippet (which has been extended) can be triggered rather than an explicit upgrade script?

src/test/java/org/folio/migration/FeeFineTypesMigrationTest.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.concurrent.CompletableFuture;
+
+import org.folio.rest.persist.PostgresClient;
+import org.folio.rest.tools.PomReader;
+import org.folio.test.support.ApiTests;
+import org.junit.Test;
+
+import io.vertx.core.Vertx;
+
+public class FeeFineTypesMigrationTest extends ApiTests {
```
marcjohnson-kint, line 16: The script doesn't appear to be a regular upgrade script, more like the loading of default reference data.

Maybe it is worth naming this test after that purpose instead.

```suggestion
public class FeeFineTypesDefaultReferenceRecordsTest extends ApiTests {
```

src/test/java/org/folio/rest/impl/TenantRefAPITest.java
```diff
@@ -92,47 +86,10 @@ public void shouldFailIfNoOkapiUrlHeaderSpecified(TestContext context) {
...
-        for (int i = 0; i < expectedFeeFines.size(); i++) {
-          final Feefine expected = expectedFeeFines.get(i);
-          final Feefine actual = createdFeeFines.get(i);
-
-          assertEquals(expected.getId(), actual.getId());
-          assertEquals(expected.getFeeFineType(), actual.getFeeFineType());
-          assertEquals(expected.getAutomatic(), actual.getAutomatic());
-        }
-        return context;
-      }).onComplete(context.asyncAssertSuccess());
```
marcjohnson-kint, line 135: Thank you for splitting this out into a separate test

src/main/resources/templates/db_scripts/schema.json
bohdan-suprun: Yes, the snippet won't be executed if we don't change the version. But you're right, I'll create a separate script.

src/test/java/org/folio/migration/FeeFineTypesMigrationTest.java
bohdan-suprun: Make sense, renamed.

src/main/resources/templates/db_scripts/schema.json
marcjohnson-kint: I think the confusion stems from our tooling having two different ways to do this and no clear guidance as to which to use

marcjohnson-kint (approved):

PR #139, opened 2020-10-28
[MODFEE-88] UUID fields should be validated against the UUID regex pattern

OleksandrVidinieiev:
Resolves [MODFEE-88](https://issues.folio.org/browse/MODFEE-88)

### DIFF
--- a/src/main/java/org/folio/rest/impl/AccountsAPI.java
+++ b/src/main/java/org/folio/rest/impl/AccountsAPI.java
@@ -9,7 +9,7 @@ import static org.folio.rest.domain.Action.WAIVE;
 
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
+import java.util.Objects;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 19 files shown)


src/main/java/org/folio/rest/impl/AccountsAPI.java
```diff
@@ -103,24 +103,21 @@ private CQLWrapper getCQL(String query, int limit, int offset) throws CQL2PgJSON
...
-          account.setInstanceId(instanceId);
-        });
+        accounts.stream()
+          .filter(a -> a.getItemId() != null)
+          .forEach(account ->
+            ctx.items.getItems().stream()
+              .filter(item -> item.getHoldingsRecordId() != null)
+              .filter(item -> item.getId().equals(account.getItemId()))
+              .findAny()
+              .flatMap(item -> ctx.holdings.getHoldingsRecords().stream()
```
roman-barannyk, line 113: Is that possible to extract it to separate method? From my perspective the stream inside the other stream confuses a bit

PR #141, opened 2020-11-05
MODFEE-139 - Fees/fines records in Circulation log do not contain item details

siarhei-charniak:
[MODFEE-139](https://issues.folio.org/browse/MODFEE-139) - Fees/fines records in Circulation log do not contain item details
This PR fixes an issue by adding item info into log record payload


src/test/java/org/folio/rest/impl/AccountsBulkPayWaiveTransferAPITests.java
```diff
@@ -351,6 +351,7 @@ private Account createAccount(String accountId, double amount) {
       .withId(accountId)
       .withOwnerId(randomId())
       .withUserId(USER_ID)
+      .withBarcode("barcode")
```
khandramai, line 354: Is it make sense to introduce constant BARCODE = "barcode"?

khandramai (approved):

PR #143, opened 2020-11-06
MODFEE-137: Add ability to provide custom cancellation reason for an account.

bohdan-suprun:
https://issues.folio.org/browse/MODFEE-137 - Provide ability to specify cancellation reason in accounts/:id/cancel request.

## Changes:
* Bump `feesfines` API version to `16.1`
* Update `cancelActionRequest.json` schema with new `cancellationReason` property which defaulting to `Cancelled as error` value.
* Use the reason when an account is closed and set the reason to:
  * `account.paymentStatus.name`
  * `accountAction.typeAction`
* Add a test case to verify the new behavior.

### DIFF
--- a/ramls/accounts.raml
+++ b/ramls/accounts.raml
@@ -1,6 +1,6 @@
 #%RAML 1.0
 title: Accounts
-version: v1.1
+version: v16.1
 baseUri: http://github.com/org/folio/mod-feesfines
 
 documentation:
--- a/ramls/accounts.raml
+++ b/ramls/accounts.raml
@@ -1,6 +1,6 @@
 #%RAML 1.0
 title: Accounts
-version: v1
+version: v1.1
 baseUri: http://github.com/org/folio/mod-feesfines
 
 documentation:
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -54,7 +54,7 @@
   "provides":[
     {
       "id":"feesfines",
-      "version":"16.0",
+      "version":"16.1",
       "handlers":[
         {
           "methods":[
--- a/ramls/actions/cancelActionRequest.json
+++ b/ramls/actions/cancelActionRequest.json
@@ -19,6 +19,11 @@
     "userName": {
       "type": "string",
       "description": "Name of the user that was logged in when the action was performed"
+    },
+    "cancellationReason": {
+      "type": "string",
+      "description": "Reason for cancellation",
+      "default": "Cancelled as error"
     }
   },
   "additionalProperties": false,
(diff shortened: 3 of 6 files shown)


ramls/accounts.raml
```diff
@@ -1,6 +1,6 @@
 #%RAML 1.0
 title: Accounts
-version: v1
+version: v1.1
```
OleksandrVidinieiev, line 3: All the RAML in this module have the same version since their inception - `v1`, which was never updated by previous module maintainers. Also, all the endpoings described in these RAMLs are exposed through one inteface - `feesfines`. So incremented version of the RAML will not be equal to the version of the interface in `ModuleDescriptor`. In these circumstances, do you think starting updating RAML versions now is the right thing to do?

@marcjohnson-kint I'd love to hear your thoughts too.

marcjohnson-kint, line 3: Which interface does this refer to? It is the `feesfines` interface that has changed and there does not appear to be an `accounts` interface.

marcjohnson-kint (approved):

ramls/accounts.raml
bohdan-suprun: I think yes, I'd update the RAML with 16.1 then, just to match version in the descriptor file.

OleksandrVidinieiev: But then should we update versions of all RAMLs to `v16.1` as well?

marcjohnson-kint: @OleksandrVidinieiev 

> All the RAMLs in this module have had the same version since their inception - v1, which was never updated by previous module maintainers.

This is a fairly common occurrence unfortunately, folks (including myself) often forget to change the RAML version.

> Also, all the endpoints described in these RAMLs are exposed through one interface - feesfines

My understanding is that there should be only a single RAML file for each FOLIO interface.

I think to resolve this discrepancy either
* the existing RAML files should be consolidated into a single file with version `v16.1`
* change the descriptor so it provides an interface for each of the RAML files

The latter of these would be a breaking compatibility change. The former should hopefully not change the behaviour (however I don't know that for sure, due to the way FOLIO uses code generation from RAML files).

If we don't want to do either of these at the moment, then let's at least make the RAML files reflect the single version that we currently expose in the system.

(This is likely an unfortunate side effect of how the system does use the RAML files at all, interfaces only exist in the form of entries in a descriptor)

OleksandrVidinieiev: >    the existing RAML files should be consolidated into a single file with version v16.1
>    change the descriptor so it provides an interface for each of the RAML files

Either one of these is out of scope of the story Bogdan is working on. I think I will create a tech-debt ticket to deal with this issue properly.

> If we don't want to do either of these at the moment, then let's at least make the RAML files reflect the single version that we currently expose in the system.

Yes, let's do that.

bohdan-suprun: I've update the version in `accounts.raml`. I assume the tech-debt will cover rest of the files?

OleksandrVidinieiev: Yes, thanks.

PR #145, opened 2020-11-09
MODFEE-141: Add two missing indexes

hjiebsco:
The issue surfaced during Bugfest Honeysuckle testing.


hjiebsco:
replaced with https://github.com/folio-org/mod-feesfines/pull/147

PR #154, opened 2020-12-14
MODFEE-146: Store new aged to lost settings for recalled items

K-Felk:
I added the requested fields to the RAML file for this policy type
this seems to make the build process automatically generate the
changes to the model and the underlying database.  I also inserted
the new fields in the lost item policy API test file.  Since this
file doesn't seem to check that any specific field exists in any
created policy, I didn't put in anything that specifically checks
for these fields either. However, I did confirm that that the fields
were present in the resulting posted policy by dumping the entire
response to screen and looking at it.  I also confirmed that if the
API was not expecting those fields or able to store them, it would
respond with a 422 error, which is not the case.

refs: https://issues.folio.org/browse/MODFEE-146

### DIFF
--- a/ramls/lost-item-fee-policy.json
+++ b/ramls/lost-item-fee-policy.json
@@ -22,6 +22,16 @@
       "type": "object",
       "$ref": "period.json"
     },
+    "recalledItemAgedLostOverdue": {
+      "description": "Recalled items aged to lost after overdue entered, must be 0 or > 0 and, if > 0 must have interval selected",
+      "type": "object",
+      "$ref": "period.json"
+    },
+    "patronBilledAfterRecalledItemAgedLost": {
+      "description": "Patron billed after a recalled item aged to lost entered, must be 0 or > 0 and, if > 0 must have interval selected",
+      "type": "object",
+      "$ref": "period.json"
+    },
     "chargeAmountItem": {
       "description": "Option to charge amount for item",
       "type": "object",
--- a/ramls/lost-item-fee-policy.raml
+++ b/ramls/lost-item-fee-policy.raml
@@ -1,6 +1,6 @@
 #%RAML 1.0
 title: Lost Item Fee Policies
-version: v1.1
+version: v16.3
 protocols: [ HTTP, HTTPS ]
 baseUri: http://localhost:9130
 
--- a/ramls/lost-item-fee-policy.raml
+++ b/ramls/lost-item-fee-policy.raml
@@ -1,6 +1,6 @@
 #%RAML 1.0
 title: Lost Item Fee Policies
-version: v1.0
+version: v1.1
 protocols: [ HTTP, HTTPS ]
 baseUri: http://localhost:9130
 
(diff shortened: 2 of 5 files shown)


src/test/java/org/folio/rest/impl/LostItemFeePoliciesAPITest.java
```diff
@@ -146,7 +148,10 @@ private static JsonObject createEntityJson() {
       .put("description", "This is description for undergrad standard")
       .put("itemAgedLostOverdue", new JsonObject().put("duration", 12).put("intervalId", "Months"))
       .put("patronBilledAfterAgedLost",
-        new JsonObject().put("duration", 12).put("intervalId", "Months"))
+        new JsonObject().put("duration", 5).put("intervalId", "Months"))
+      .put("recalledItemAgedLostOverdue", new JsonObject().put("duration", 12).put("intervalId", "Months"))
+      .put("patronBilledAfterRecalledItemAgedLost",
+        new JsonObject().put("duration", 6).put("intervalId", "Months"))
```
bohdan-suprun, line 154: Do we need assertion statements to verify the properties got persisted?

ramls/lost-item-fee-policy.json
```diff
@@ -22,6 +22,16 @@
       "type": "object",
       "$ref": "period.json"
     },
+    "recalledItemAgedLostOverdue": {
```
bohdan-suprun, line 25: We have to update minor version of the API that uses this schema in both RAML and Module descriptor files.

src/test/java/org/folio/rest/impl/LostItemFeePoliciesAPITest.java
K-Felk: That was my initial thought, but looking at the test, it doesn't seem to explicitly check for persistence of *any* specific attributes.  My guess based on the tinkering I did is that the original test writer assumed that if it couldn't accept an attribute, it would kick back a 422 error, so they just checked for successful creation of the policy as a whole.  If we are going to add a check for these specific attributes, we should also add a check for all of the others.

ramls/lost-item-fee-policy.json
K-Felk: Ah, of course.  I'll adjust.

src/test/java/org/folio/rest/impl/LostItemFeePoliciesAPITest.java
bohdan-suprun: Sounds good. Yeah, the API should return 422 in case of unknown property.

ramls/lost-item-fee-policy.json
K-Felk: I tried doing this-I'm not sure I did it right.  I'm noticing that the version in the module descriptor does not match the one in NEWS.md...aren't they supposed to?

bohdan-suprun: I think this are different versions. The one in the NEWS file are the module versions (from pom) but in module descriptor and in RAML it is API versions.

ramls/lost-item-fee-policy.raml
```diff
@@ -1,6 +1,6 @@
 #%RAML 1.0
 title: Lost Item Fee Policies
-version: v1.0
+version: v1.1
```
bohdan-suprun, line 3: Hi @OleksandrVidinieiev. is it fine to synchronize this version with the version from module descriptor (i.e. `16.3`)?

OleksandrVidinieiev: Yes. We were going to do that eventually anyway.

K-Felk: Okay, so instead of version 1.1, this should be changed to 16.3?

bohdan-suprun: Thanks! @K-Felk, we can set 16.3 here.

bohdan-suprun (approved):

PR #158, opened 2021-01-21
MODFEE-164 Refund report fails when automatic refunds exist

Illia-Dolzhenko:
Resolves: [MODFEE-164](https://issues.folio.org/browse/MODFEE-164)

### DIFF
--- a/src/main/java/org/folio/rest/service/report/RefundReportService.java
+++ b/src/main/java/org/folio/rest/service/report/RefundReportService.java
@@ -464,10 +464,16 @@ public class RefundReportService {
   }
 
   private static String getStaffInfo(String comments) {
+    if(comments == null){
+      return "";
+    }
     return parseFeeFineComments(comments).get(STAFF_COMMENTS_KEY);
   }
 
   private static String getPatronInfo(String comments) {
+    if(comments == null){
+      return "";
+    }
     return parseFeeFineComments(comments).get(PATRON_COMMENTS_KEY);
   }
 
--- a/src/main/java/org/folio/rest/service/report/RefundReportService.java
+++ b/src/main/java/org/folio/rest/service/report/RefundReportService.java
@@ -24,6 +24,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 
+import org.apache.commons.lang.StringUtils;
 import org.folio.rest.client.ConfigurationClient;
 import org.folio.rest.client.InventoryClient;
 import org.folio.rest.client.UserGroupsClient;
@@ -464,17 +465,11 @@ public class RefundReportService {
   }
 
   private static String getStaffInfo(String comments) {
-    if(comments == null){
-      return "";
-    }
-    return parseFeeFineComments(comments).get(STAFF_COMMENTS_KEY);
+    return StringUtils.defaultString(parseFeeFineComments(StringUtils.defaultString(comments)).get(STAFF_COMMENTS_KEY));
   }
 
   private static String getPatronInfo(String comments) {
-    if(comments == null){
-      return "";
-    }
-    return parseFeeFineComments(comments).get(PATRON_COMMENTS_KEY);
+    return StringUtils.defaultString(parseFeeFineComments(StringUtils.defaultString(comments)).get(PATRON_COMMENTS_KEY));
   }
 
   private static String getItemBarcode(RefundReportContext ctx, String accountId) {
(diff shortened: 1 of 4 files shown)


src/test/java/org/folio/rest/impl/FeeFineReportsAPITest.java
```diff
@@ -544,6 +561,12 @@ private Feefineaction createAction(int actionCounter, Account account, String da
       staffInfo, patronInfo, txInfo);
   }
 
+  private Feefineaction createActionWithoutComments(Account account, String dateTime,
+                                     String type, String method, Double amount, Double balance, String txInfo) {
```
alexanderkurash, line 565: Something is wrong with indentation here

src/test/java/org/folio/rest/impl/FeeFineReportsAPITest.java
```diff
@@ -558,6 +581,18 @@ private Feefineaction createAction(String userId, int actionCounter, Account acc
     return action;
   }
 
+  private Feefineaction createActionWithoutComments(String userId, Account account, String dateTime,
+                                     String type, String method, Double amount, Double balance, String txInfo) {
```
alexanderkurash, line 585: Same here

src/test/java/org/folio/test/support/EntityBuilder.java
```diff
@@ -117,6 +117,20 @@ public static Feefineaction buildFeeFineAction(String userId, String accountId,
       .withComments(format("STAFF : %s \n PATRON : %s", commentForStaff, commentForPatron));
   }
 
+  public static Feefineaction buildFeeFineActionWithoutComments(String userId, String accountId, String type,
+                                                                String paymentMethod, Double amount, Double balance, Date date) {
```
alexanderkurash, line 121: And here

src/test/java/org/folio/rest/impl/FeeFineReportsAPITest.java
```diff
@@ -544,6 +561,12 @@ private Feefineaction createAction(int actionCounter, Account account, String da
       staffInfo, patronInfo, txInfo);
   }
 
+  private Feefineaction createActionWithoutComments(Account account, String dateTime,
```
alexanderkurash, line 564: I think something like `createActionWithNullComments` would be more explicit. `without` doesn't really tell much - is it empty string, null or maybe the field is missing.

src/main/java/org/folio/rest/service/report/RefundReportService.java
```diff
@@ -464,10 +464,16 @@ private static String singleOrDefaultMessage(List<String> values, String message
   }
 
   private static String getStaffInfo(String comments) {
+    if(comments == null){
```
alexanderkurash, line 467: Just to preserve our existing code style, can we add spaces here? Like in `if (condition()) {`

OleksandrVidinieiev, line 467: You can try something like `StringUtils.defaultString(String s)` instead

src/main/java/org/folio/rest/service/report/RefundReportService.java
```diff
@@ -464,11 +465,11 @@ private static String singleOrDefaultMessage(List<String> values, String message
   }
 
   private static String getStaffInfo(String comments) {
-    return parseFeeFineComments(comments).get(STAFF_COMMENTS_KEY);
+    return StringUtils.defaultString(parseFeeFineComments(StringUtils.defaultString(comments)).get(STAFF_COMMENTS_KEY));
```
OleksandrVidinieiev, line 468: Maybe `parseFeeFineComments` should check if `comments` is null and return an empty map if it is? This way we can avoid second call of `defaultString` here. And please add static imports for utility methods where it makes sense.

src/test/java/org/folio/rest/impl/FeeFineReportsAPITest.java
```diff
@@ -558,6 +581,18 @@ private Feefineaction createAction(String userId, int actionCounter, Account acc
     return action;
   }
 
+  private Feefineaction createActionWithNullComments(String userId, Account account, String dateTime,
```
OleksandrVidinieiev, line 584: Do we really need two `createActionWithNullComments` methods? You used only one of them in test...

src/main/java/org/folio/rest/service/report/RefundReportService.java
```diff
@@ -464,11 +465,11 @@ private static String singleOrDefaultMessage(List<String> values, String message
   }
 
   private static String getStaffInfo(String comments) {
```
OleksandrVidinieiev, line 467: What happens if we just return `null` from this method? Will it break anything? I'm just thinking if we really need this `defaultString` call here...

Illia-Dolzhenko: It will break RefundReportEntryMatcher and will remove staffInfo and patronInfo fields from json for ReportEntry

OleksandrVidinieiev (approved):

PR #160, opened 2021-01-21
MODFEE-150 Updated account dateUpdated field on fees/fines actions

Roman-Leshchenko:


src/main/java/org/folio/rest/service/action/ActionService.java
```diff
@@ -148,6 +148,9 @@ protected Feefineaction createFeeFineActionAndUpdateAccount(Account account, Mon
       .withDateAction(new Date());
 
     account.getPaymentStatus().setName(actionType);
+    account.setDateUpdated(new Date());
+    account.getMetadata().setUpdatedDate(new Date());
```
alexanderkurash, line 152: Do you know which one of these fields is used on FE for "Date updated" column? Also, I though that `metadata` fields should be taken care of by RMB, but apparently they aren't?

src/test/java/org/folio/test/support/matcher/AccountMatchers.java
```diff
@@ -21,4 +28,30 @@ private AccountMatchers() {}
...
+
+    final Account updatedAccount = getAccountByIdResponse.as(Account.class);
+    final Date dateCreated = updatedAccount.getMetadata().getCreatedDate();
+    final Date dateUpdated = updatedAccount.getMetadata().getUpdatedDate();
+
+    assertThat(dateUpdated, equalTo(updatedAccount.getDateUpdated()));
+    assertThat(dateCreated, not(equalTo(dateUpdated)));
+
+    return updatedAccount;
+  }
```
alexanderkurash, line 56: I like the idea of moving this logic to a single place and using it from different test classes, but `...Matcher` class is probably not the best place for it. Our matcher classes typically just have static methods that create JSON matchers, they don't know anything about clients or making requests etc. To me this looks like more of a `...Test` class concern. What if we create a class that extends `ApiTests`, put this method in it and make all the `...Test` classes that need this method to extend it? Or move this method to a new util class?

src/main/java/org/folio/rest/service/action/ActionService.java
Roman-Leshchenko: FE uses `metedata: updatedDate`. 
It will be little bit weird if  an account is updated and only `metedata: updatedDate` will be changed. In this case account  dateUpdated seems uncessary and duplicates "metedata: updatedDate".

Account is updated via PostgresClient, that's why RMB doesn't update `metadata` . Or RMB  should handle it too?

src/test/java/org/folio/test/support/matcher/AccountMatchers.java
Roman-Leshchenko: Agree with you. I like the approach with `BaseTest`.

src/test/java/org/folio/test/support/AccountsActionAPITests.java
```diff
@@ -0,0 +1,41 @@
...
+import static org.hamcrest.CoreMatchers.not;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.notNullValue;
+
+import io.restassured.response.Response;
+import java.util.Date;
+import org.folio.rest.jaxrs.model.Account;
+import org.folio.rest.utils.ResourceClient;
+
+public abstract class AccountsActionAPITests extends ApiTests {
```
alexanderkurash, line 14: Maybe we can just call it `ActionsAPITests`. It would be consistent with our names for the "new" actions functionality - class `Action`, for example.

src/main/java/org/folio/rest/service/action/ActionService.java
OleksandrVidinieiev: Yes, it should. See `schema.json`:
```
      "tableName":"accounts",
      "fromModuleVersion":"15.0",
      "withMetadata":true
```

I don't think `Account.dateUpdated` was ever used. At least I could not find a single account instance with this property. We should not remove it from the schema though, as technically it would be a breaking change.

src/test/java/org/folio/test/support/AccountsActionAPITests.java
Roman-Leshchenko: It makes sense.  I will change the name

src/test/java/org/folio/test/support/AccountsActionAPITests.java
```diff
@@ -0,0 +1,41 @@
...
+
+  public static Account verifyAccountAndGet(
+    ResourceClient accountsClient,
+    String accountId,
+    String expectedPaymentStatus,
+    float amount,
```
OleksandrVidinieiev, line 20: Can we accept `amount` as a `double` and convert it to `float` right before comparison? Just so we could test values retrieved directly from JSONs without casting them every time.

src/main/java/org/folio/rest/service/action/ActionService.java
Roman-Leshchenko: According  [RMB  doc](https://github.com/folio-org/raml-module-builder#metadata) the `metadata` is updated by [triggers](https://github.com/folio-org/raml-module-builder/blob/master/domain-models-runtime/src/main/resources/templates/db_scripts/metadata.ftl). However only `created...` fields is hadled by the trigger.

alexanderkurash: I wonder if this is a bug in RMB. In the doc it says that "RMB will set both update and create values to the same date/time and to the same user"

src/main/java/org/folio/rest/service/action/ActionService.java
```diff
@@ -148,6 +148,9 @@ protected Feefineaction createFeeFineActionAndUpdateAccount(Account account, Mon
       .withDateAction(new Date());
 
     account.getPaymentStatus().setName(actionType);
+    account.setDateUpdated(new Date());
```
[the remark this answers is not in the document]
alexanderkurash: Well, it looks like no one has ever used `dateUpdated`. The UI was looking at `metadata.updatedDate`, which is supposed to be populated by RMB, but for some reason it isn't.

src/main/java/org/folio/rest/service/action/ActionService.java
Roman-Leshchenko: I would like to retain these changes in PR. 

Removing `Account.setDateUpdated`  can be confusing in the future.
`Metadata..setUpdatedDate`  actually fixes the [issue](https://issues.folio.org/browse/MODFEE-150).  

As for RMB we can create ticket to create DB trigger on update action.

src/main/java/org/folio/rest/service/action/ActionService.java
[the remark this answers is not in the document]
mtraneis: Thanks for explaining, @alexanderkurash.  If we make this change we will also change the UI logic to use `metadata.updatedDate` as well?  If RMB is the problem should we also file a bug on RMB?

src/main/java/org/folio/rest/service/action/ActionService.java
OleksandrVidinieiev: Is this the only place in the code where we update an account? If it's not, we should update `Account.dateUpdated` in all those places too.

src/main/java/org/folio/rest/service/action/ActionService.java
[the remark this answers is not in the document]
Roman-Leshchenko: `metadata.updatedDate`  is already used by frontend, no changes are required.

src/main/java/org/folio/rest/service/action/ActionService.java
Roman-Leshchenko: I will go with your suggestion.

OleksandrVidinieiev: I just tried debugging the whole thing. Looks like metadata is populated in an entity before it event gets to the implementation of a generated interface (*API class). When RMB converts JSON-body of the request into a POJO entity, it checks if entity has a `setMetadata()` method. And if it does, RMB uses this method to populate metadata.

Since in this case we did not receive `Account` in the request, its metadata is never updated. And for the same reason none of our `FeeFineActions` contain metadata.

RMD uses this method internally to populate metadata:
```
MetadataUtil.populateMetadata(Object entity, Map<String, String> okapiHeaders)
```
You can use it too to fix your problem. Just use headers you pass into the constructor of `ActionService` as second argument.

Roman-Leshchenko: @OleksandrVidinieiev  Thanks, nice find!

src/main/java/org/folio/rest/service/action/ActionService.java
```diff
@@ -159,6 +164,15 @@ protected Feefineaction createFeeFineActionAndUpdateAccount(Account account, Mon
     return feeFineAction;
   }
 
+  private void populateMetadata(Account account) {
+    try {
+      MetadataUtil.populateMetadata(account, headers);
+    } catch (ReflectiveOperationException e) {
+      log.error("Can not populate Metadata, cause {}", e.getMessage());
```
OleksandrVidinieiev, line 171: Please mention object type ("account") and it's ID

src/main/java/org/folio/rest/service/action/ActionService.java
```diff
@@ -159,6 +164,15 @@ protected Feefineaction createFeeFineActionAndUpdateAccount(Account account, Mon
...
+      account.getMetadata().setUpdatedDate(new Date());
```
OleksandrVidinieiev, line 172: I think it is possible that metadata is null. For example, if an account was created not from a JSON-request, but manually from the code. So please take necessary precautions.

alexanderkurash (approved):

PR #162, opened 2021-01-22
MODFEE-165 DELETE feefineactions endpoint is broken

alexanderkurash:
Resolves [MODFEE-165](https://issues.folio.org/browse/MODFEE-165).

Field name was missing from the criteria for deleting fee/fine actions.


OleksandrVidinieiev (approved):

src/main/java/org/folio/rest/impl/FeeFineActionsAPI.java
```diff
@@ -44,6 +44,7 @@
 public class FeeFineActionsAPI implements Feefineactions {
 
   private static final String FEEFINEACTIONS_TABLE = "feefineactions";
+  private static final String FEEFINEACTIONS_ID_FIELD = "'id'";
```
roman-barannyk, line 47: It seems like we already have the constant with the same value?

roman-barannyk (approved):

PR #163, opened 2021-01-22
MODFEE-160 Add Fee/fine owner as criteria for 'Refunds to process manually report'

roman-barannyk:
## Purpose
- Add Fee/fine owner as criteria for 'Refunds to process manually report'

Resolves: [MODFEE-160](https://issues.folio.org/browse/MODFEE-160)

## Approach
- Change method for "/feefine-reports/refund" from GET to POST;
- Add ownerId parameters;
- Update existing test;
- Create new test;

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -933,15 +933,15 @@
     },
     {
       "id" : "feefine-reports",
-      "version" : "1.0",
+      "version" : "2.0",
       "handlers": [
         {
           "methods":[
-            "GET"
+            "POST"
           ],
           "pathPattern":"/feefine-reports/refund",
           "permissionsRequired":[
-            "feefine-reports.refund.get"
+            "feefine-reports.refund.post"
           ],
           "modulePermissions": [
             "configuration.entries.collection.get",
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -195,8 +195,10 @@
             "DELETE"
           ],
           "pathPattern":"/accounts/{id}",
-          "permissionsRequired":[
-            "accounts.item.delete",
+          "permissionsRequired": [
+            "accounts.item.delete"
+          ],
+          "modulePermissions": [
             "pubsub.publish.post"
           ]
         },
--- a/src/main/java/org/folio/rest/impl/FeeFineReportsAPI.java
+++ b/src/main/java/org/folio/rest/impl/FeeFineReportsAPI.java
@@ -7,6 +7,7 @@ import java.util.Map;
 
 import javax.ws.rs.core.Response;
 
+import org.folio.rest.annotations.Validate;
 import org.folio.rest.exception.FailedValidationException;
 import org.folio.rest.jaxrs.model.RefundReport;
 import org.folio.rest.jaxrs.model.RefundReportRequest;
@@ -28,6 +29,7 @@ public class FeeFineReportsAPI implements FeefineReports {
     "Invalid startDate or endDate parameter";
   private static final String INTERNAL_SERVER_ERROR_MESSAGE = "Internal server error";
 
+  @Validate
   @Override
   public void postFeefineReportsRefund(RefundReportRequest entity, Map<String,
     String> okapiHeaders, Handler<AsyncResult<Response>> asyncResultHandler,
--- a/src/main/java/org/folio/rest/impl/FeeFineReportsAPI.java
+++ b/src/main/java/org/folio/rest/impl/FeeFineReportsAPI.java
@@ -9,9 +9,9 @@ import javax.ws.rs.core.Response;
 
 import org.folio.rest.exception.FailedValidationException;
 import org.folio.rest.jaxrs.model.RefundReport;
+import org.folio.rest.jaxrs.model.RefundReportRequest;
 import org.folio.rest.jaxrs.resource.FeefineReports;
 import org.folio.rest.service.report.RefundReportService;
-import org.folio.util.UuidUtil;
 import org.joda.time.DateTime;
 import org.joda.time.format.ISODateTimeFormat;
 
@@ -30,23 +30,15 @@ public class FeeFineReportsAPI implements FeefineReports {
   private static final String INTERNAL_SERVER_ERROR_MESSAGE = "Internal server error";
 
   @Override
-  public void getFeefineReportsRefund(String startDate, String endDate, String ownerId,
-    Map<String, String> okapiHeaders, Handler<AsyncResult<Response>> asyncResultHandler,
+  public void postFeefineReportsRefund(RefundReportRequest entity, Map<String,
+    String> okapiHeaders, Handler<AsyncResult<Response>> asyncResultHandler,
     Context vertxContext) {
 
+    DateTime startDate = parseDate(entity.getStartDate());
+    DateTime endDate = parseDate(entity.getEndDate());
     log.info("Refund report requested, parameters: startDate={}, endDate={}", startDate, endDate);
 
-    DateTime startDateTime = parseDate(startDate);
-    DateTime endDateTime = parseDate(endDate);
-    if (ownerId != null && !UuidUtil.isUuid(ownerId)) {
-      log.error("Invalid ownerId={}", ownerId);
-
-      handleRefundReportResult(
-        failedFuture(new FailedValidationException(INVALID_OWNER_ID_MESSAGE)),
-        asyncResultHandler);
-    }
-
-    if (startDateTime == null || endDateTime == null) {
+    if (startDate == null || endDate == null) {
       log.error("Invalid parameters: startDate={}, endDate={}", startDate, endDate);
 
       handleRefundReportResult(
@@ -54,7 +46,7 @@ public class FeeFineReportsAPI implements FeefineReports {
         asyncResultHandler);
     } else {
       new RefundReportService(okapiHeaders, vertxContext)
-        .buildReport(startDateTime, endDateTime, ownerId)
+        .buildReport(startDate, endDate, entity.getFeeFineOwners())
         .onComplete(result -> handleRefundReportResult(result, asyncResultHandler));
     }
   }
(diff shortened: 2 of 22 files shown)


src/main/java/org/folio/rest/repository/FeeFineActionRepository.java
```diff
@@ -180,6 +182,16 @@ private GroupedCriterias getTypeCriteria(Action action) {
...
+    return accountIds.stream()
+      .map(accountId -> new Criteria()
+        .addField("'accountId'")
+        .setOperation("=")
+        .setVal(accountId)
+        .setJSONB(true))
+      .collect(Collectors.toList());
+  }
+
   private GroupedCriterias groupCriterias(List<Criteria> criterias, String op) {
```
mtraneis, line 195: Can this method be refactored into a common utility class of some sort so that it can be reused by AccountRepository.java, which recreates this method for its use?

src/test/java/org/folio/rest/impl/FeeFineReportsAPITest.java
```diff
@@ -189,14 +192,12 @@ public void emptyReportWhenRefundedAfterEndDate() {
...
   public void badRequestWhenParameterIsMissingOrMalformed() {
-    refundReportsClient.getByParameters("nonExistingParam=1", HTTP_BAD_REQUEST);
-    refundReportsClient.getByParameters("startDate=2020-01-01", HTTP_BAD_REQUEST);
-    refundReportsClient.getByParameters("endDate=2020-01-01", HTTP_BAD_REQUEST);
-    refundReportsClient.getByParameters("startDate=not-a-date", HTTP_BAD_REQUEST);
-    refundReportsClient.getByParameters("endDate=not-a-date", HTTP_BAD_REQUEST);
-    refundReportsClient.getByParameters("startDate=2020-01-01&endDate=not-a-date", HTTP_BAD_REQUEST);
-    refundReportsClient.getByParameters("startDate=not-a-date&endDate=2020-01-01", HTTP_BAD_REQUEST);
-    refundReportsClient.getByParameters("startDate=not-a-date&endDate=not-a-date", HTTP_BAD_REQUEST);
+    refundReportsClient.getFeeFineRefundReports("nonExistingParam=1", "2020-01-01", HTTP_UNPROCESSABLE_ENTITY);
```
mtraneis, line 195: I think it's arguable that the server should return 400 when it receives malformed arguments such as these.

descriptors/ModuleDescriptor-template.json
```diff
@@ -933,15 +933,15 @@
     },
     {
       "id" : "feefine-reports",
-      "version" : "1.0",
+      "version" : "2.0",
       "handlers": [
         {
           "methods":[
-            "GET"
+            "POST"
```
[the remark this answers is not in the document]
roman-barannyk: yes, it was done to avoid URL characters restrictions. We decided to not limit ownerIds in case there are more then 20 of them on production environment. These changes were discussed with FE team and UI part will support POST

src/main/java/org/folio/rest/repository/FeeFineActionRepository.java
roman-barannyk: I have already done some refactoring in the code and it was removed

src/test/java/org/folio/rest/impl/FeeFineReportsAPITest.java
roman-barannyk: But I guess if the body was built correctly and there is only semantical error for startDate value is should be 422... Am I wrong?

src/main/java/org/folio/rest/repository/AccountRepository.java
```diff
@@ -4,11 +4,16 @@
...
+import java.util.stream.Collectors;
 
 import org.folio.rest.domain.FeeFineNoticeContext;
 import org.folio.rest.jaxrs.model.Account;
 import org.folio.rest.jaxrs.model.Feefineaction;
+import org.folio.rest.persist.Criteria.Criteria;
+import org.folio.rest.persist.Criteria.Criterion;
+import org.folio.rest.persist.Criteria.GroupedCriterias;
 import org.folio.rest.persist.PostgresClient;
+import org.folio.rest.persist.interfaces.Results;
```
roman-barannyk, line 16: remove unused imports please

src/main/java/org/folio/rest/repository/FeeFineActionRepository.java
```diff
@@ -132,16 +131,39 @@ public FeeFineActionRepository(Map<String, String> headers, Context context) {
...
-    return this.get(criterion);
+  public Future<List<Feefineaction>> findActionsByTypeForPeriodAndOwners(Action typeAction,
+    String startDate, String endDate, List<String> ownerIds, int limit) {
+
+    String ownerIdsFilter = ownerIds != null && !ownerIds.isEmpty()
+      ? String.format("AND accounts.jsonb->>'ownerId' IN (%s)",
+      ownerIds.stream()
+        .map(id -> format("'%s'", id))
+        .collect(Collectors.joining(",")))
+      : EMPTY;
```
roman-barannyk, line 142: extract it to separate method, for example: buildOwnerIdsFilter()

src/main/java/org/folio/rest/repository/FeeFineActionRepository.java
```diff
@@ -132,16 +131,39 @@ public FeeFineActionRepository(Map<String, String> headers, Context context) {
...
+    String typeActions = List.of(
+      typeAction.getFullResult(), typeAction.getPartialResult()).stream()
+      .map(result -> format("'%s'", result))
+      .collect(Collectors.joining(","));
```
roman-barannyk, line 146: try to simplify it in SQL request

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -40,6 +40,10 @@
         {
           "fieldName" : "loanId",
           "tOps" : "ADD"
+        },
+        {
+          "fieldName" : "ownerId",
+          "tOps" : "ADD"
```
roman-barannyk, line 46: We can do it in scope of another ticket for performance testing

src/test/java/org/folio/rest/impl/FeeFineReportsAPITest.java
```diff
@@ -519,13 +582,29 @@ private void requestAndCheck(List<RefundReportEntry> reportEntries) {
...
+
+    ValidatableResponse response = requestRefundReport(START_DATE, END_DATE, ownerIds)
+      .then()
+      .statusCode(HttpStatus.SC_OK)
+      .body("reportData", iterableWithSize(reportEntries.size()));
+
+    IntStream.range(0, reportEntries.size())
+      .forEach(index -> response.body(format("reportData[%d]", index),
+        refundReportEntryMatcher(reportEntries.get(index))));
+  }
```
roman-barannyk, line 596: Try to avoid code duplicate

src/test/java/org/folio/rest/utils/ReportResourceClient.java
```diff
@@ -1,30 +1,59 @@
...
+
+  public Response getFeeFineRefundReports(String startDate, String endDate,
+    HttpStatus expectedStatus) {
+
+    return okapiClient.post(baseUri, createRefundReportRequest(startDate, endDate, null))
       .then()
       .statusCode(expectedStatus.toInt())
       .extract()
       .response();
   }
```
roman-barannyk, line 43: remove it and call getFeeFineRefundReports(String startDate, String endDate, List<String> ownerIds,
    HttpStatus expectedStatus) with null value for owners

src/test/java/org/folio/rest/utils/ReportResourceClient.java
```diff
@@ -1,30 +1,59 @@
...
+
+  private String createRefundReportRequest(String startDate, String endDate,
+    List<String> ownerIds) {
+
+    JsonArray feeFineOwners = new JsonArray();
+    if (ownerIds != null) {
+      ownerIds.forEach(feeFineOwners::add);
```
roman-barannyk, line 50: we can use constructor for JsonArray

src/main/java/org/folio/rest/impl/FeeFineReportsAPI.java
```diff
@@ -28,43 +29,43 @@
...
         .respond200WithApplicationJson(asyncResult.result())));
     }
     else if (asyncResult.failed()) {
       final Throwable cause = asyncResult.cause();
       if (cause instanceof FailedValidationException) {
         log.error("Report parameters validation failed: " + cause.getLocalizedMessage());
-        asyncResultHandler.handle(succeededFuture(FeefineReports.GetFeefineReportsRefundResponse
-          .respond400WithTextPlain(cause.getLocalizedMessage())));
+        asyncResultHandler.handle(succeededFuture(FeefineReports.PostFeefineReportsRefundResponse
+          .respond422WithTextPlain(cause.getLocalizedMessage())));
```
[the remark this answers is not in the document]
roman-barannyk: Now RMB validates request according to the schema. But it returns also 422 even if incorrect/malformed syntax in the request.

src/test/java/org/folio/rest/impl/FeeFineReportsAPITest.java
roman-barannyk: I have renamed this test and removed some redundant cases

descriptors/ModuleDescriptor-template.json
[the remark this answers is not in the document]
mtraneis: Ok, if you think it's necessary to use POST then this change should be kept. Sometimes we programmers account for all cases that might never be used, and it's better to avoid changing the API drastically as much as possible, especially in this case when we have to use POST to do a GET.  IMHO even if they would send in 40 UUIDs there would still be room for it, and 20 is a practical maximum, so a GET is good enough, but since there were prior discussions I trust your judgment.

src/main/java/org/folio/rest/impl/FeeFineReportsAPI.java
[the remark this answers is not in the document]
mtraneis: Sounds good. As long as syntax errors are covered by RMB then we don't need to cover it here.

src/test/java/org/folio/rest/impl/FeeFineReportsAPITest.java
mtraneis: After removing the invalid syntax cases from this test there is no reason to return a 400 here.

alexanderkurash (approved):

PR #165, opened 2021-02-09
MODFEE-171 Remove end date from required

Illia-Dolzhenko:
Resolves: [MODFEE-171](https://issues.folio.org/browse/MODFEE-171)


alexanderkurash:
Will be implemented as a part of MODFEE-173

PR #168, opened 2021-02-19
MODFEE-169: Add missing indexes

mtraneis:
Added two missing b-tree indexes for the field "userId" in the accounts and manualblocks tables.
I did not add an index for accounts.jsonb-->'status', as specified in the JIRA, because it is an enum table and only has a few values.

### DIFF
--- a/src/main/resources/templates/db_scripts/schema.json
+++ b/src/main/resources/templates/db_scripts/schema.json
@@ -34,12 +34,16 @@
     },
     {
       "tableName":"accounts",
-      "fromModuleVersion":"15.0",
+      "fromModuleVersion":"15.9.2",
       "withMetadata":true,
       "index" : [
         {
           "fieldName" : "loanId",
           "tOps" : "ADD"
+        },
+        {
+          "fieldName": "userId",
+          "tOps": "ADD"
         }
       ]
     },
@@ -81,8 +85,14 @@
     },
     {
       "tableName":"manualblocks",
-      "fromModuleVersion":"15.0",
-      "withMetadata":true
+      "fromModuleVersion":"15.9.2",
+      "withMetadata":true,
+      "index" : [
+        {
+          "fieldName" : "userId",
+          "tOps" : "ADD"
+        }
+      ]
     },
     {
       "tableName":"manual_block_templates",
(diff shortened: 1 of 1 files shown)


src/main/resources/templates/db_scripts/schema.json
```diff
@@ -34,12 +34,16 @@
     },
     {
       "tableName":"accounts",
-      "fromModuleVersion":"15.0",
+      "fromModuleVersion":"15.9.2",
```
alexanderkurash, line 37: According to RMB doc, it looks like fromModuleVersion should be 15.10.0, otherwise we won't get this new index when upgrading from 15.9.2 to 15.10.0

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -34,12 +34,16 @@
...
       "withMetadata":true,
       "index" : [
         {
           "fieldName" : "loanId",
           "tOps" : "ADD"
+        },
+        {
+          "fieldName": "userId",
```
alexanderkurash, line 45: Will this actually work for the query
lower(f_unaccent(accounts.jsonb->>'userId')) LIKE lower(f_unaccent('e96618a9-04ee-4fea-aa60-306a8f4dd89b'))
Does the warning actually disappear with this index?

mtraneis: I would think so. I haven't tested it, but we did a similar thing for other indexes with the same type of warning.  I can try to test it if you want.

src/main/resources/templates/db_scripts/schema.json
mtraneis: Thanks. I didn't really know what this was about. Updated it.

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -81,8 +85,14 @@
     },
     {
       "tableName":"manualblocks",
-      "fromModuleVersion":"15.0",
-      "withMetadata":true
+      "fromModuleVersion":"15.9.2",
```
alexanderkurash, line 88: Should we also change a version here?

mtraneis: Done

alexanderkurash (approved):

PR #179, opened 2021-04-27
MODFEE-187 - Search/filter on fees/fines not pulling recent circ log actions

VitaliiZahoruiko:
JIRA issue: https://issues.folio.org/browse/MODFEE-187

### Purpose
When creating new circulation action, like fees/fines - refunded partially, filtering by user barcode, item barcode or fee/fine drop down filters don't show actions

### Approach
- Changed Refund and Credit actions to be processed by logEventService and published as FeeFineLogEvent
- Changed tests

### DIFF
--- a/src/main/java/org/folio/rest/impl/FeeFineActionsAPI.java
+++ b/src/main/java/org/folio/rest/impl/FeeFineActionsAPI.java
@@ -2,7 +2,6 @@ package org.folio.rest.impl;
 
 import static org.apache.commons.lang3.BooleanUtils.isTrue;
 import static org.folio.rest.domain.Action.CREDIT;
-import static org.folio.rest.domain.Action.REFUND;
 import static org.folio.rest.service.LogEventPublisher.LogEventPayloadType.FEE_FINE;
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/folio/rest/impl/FeeFineActionsAPI.java
```diff
@@ -164,20 +163,14 @@ private Response sendPatronNoticeIfNeedBe(Feefineaction entity, Map<String, Stri
 
   private Future<Response> publishLogEvent(Feefineaction entity, Map<String, String> okapiHeaders,
     Context vertxContext, Response response) {
-
-    // do not publish log records for CREDIT and REFUND actions
-    if (!CREDIT.isActionForResult(entity.getTypeAction()) &&
```
khandramai, line 169: Why was this if-block initially presented? Does this conflict with the original business requirements statement? CC @siarhei-charniak

khandramai (approved):

alekGbuz (approved):

siarhei-charniak (approved):

PR #181, opened 2021-05-26
MODFEE-189: Lost item fees (closed) are not cancelled

ymacedor:
When resolving a claimed returned situation, lost item fees (closed) are not cancelled.

https://issues.folio.org/browse/MODFEE-189


alexanderkurash:
After discussion with Holly Mistlebauer, it looks like a better approach here would be to make an account to go back to `Open` status during step 5 from the Jira ticket (_Claim the item as returned in Loan Details, which will refund the 100.00 Lost Item Process Fee and suspend both fees_) rather than make an exception in the rule that we don't allow to cancel `Closed` accounts. If we do that, both `Lost Item Fee` and `Lost Item Processing Fee` would be in status `Open` when we get to step 6, so they both would be cancelled without issues. It also makes more sense from the business logic standpoint.

ymacedor:
Thanks @alexanderkurash,  that must work, we will implement your suggestion in the frontend.

ymacedor:
This issue was solved in the frontend

PR #187, opened 2021-06-17
[MODFEE-98] Add UUID regular expression pattern to all id properties in JSON schemas

Mykyta-Varenyk:
Add UUID regular expression pattern to all id properties in JSON schemas.
Resolves : https://issues.folio.org/browse/MODFEE-98.
mod-feesfines is required for mod-patron and mod-circulation, and is optional for mod-user-bl.
Regarding ui modules, the following modules are dependent on feesfines: ui-plugin-bursas-export,ui-circulation,ui-users.
Please don't merge before other modules are updated.
There is an issue with feefineactiondata.json, where createdAt property has already records in the database with different value types -> UUID and string.

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -54,7 +54,7 @@
   "provides":[
     {
       "id":"feesfines",
-      "version":"16.3",
+      "version":"16.4",
       "handlers":[
         {
           "methods":[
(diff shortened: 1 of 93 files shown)


roman-barannyk:
It seems like the interface version should be upgraded as well...

descriptors/ModuleDescriptor-template.json
```diff
@@ -54,7 +54,7 @@
   "provides":[
     {
       "id":"feesfines",
-      "version":"16.3",
+      "version":"16.4",
```
OleksandrVidinieiev, line 57: Update major version

Mykyta-Varenyk: Resolved

OleksandrVidinieiev (approved):

PR #188, opened 2021-06-22
Modfee 29

nikolai991:


pom.xml
```diff
@@ -136,6 +136,11 @@
       <version>1.1.1</version>
       <scope>test</scope>
     </dependency>
+      <dependency>
+          <groupId>org.folio</groupId>
+          <artifactId>mod-feesfines</artifactId>
+          <version>15.9.3-SNAPSHOT</version>
+      </dependency>
```
OleksandrVidinieiev, line 143: Why should the module depend on itself?

ramls/inventory/holdingsrecord.json
```diff
@@ -35,6 +36,7 @@
       "$ref" : "../uuid.json"
     },
     "permanentLocation": {
+      "description": "Inventory instances identifier",
```
alexanderkurash, line 39: Wrong description here? 

Also, do we actually need to add these descriptions in scope of MODFEE-29?

src/test/java/org/folio/rest/impl/AccountsRefundAPITests.java
```diff
@@ -83,6 +83,27 @@
   private final ResourceClient refundClient = buildAccountsRefundClient(FIRST_ACCOUNT_ID);
   private final ResourceClient bulkRefundClient = buildAccountBulkRefundClient();
 
+  private final MonetaryValue[]  BALANCE_MINUS = {
```
alexanderkurash, line 86: BALANCE_NEGATIVE?

src/test/java/org/folio/rest/impl/AccountsRefundAPITests.java
```diff
@@ -83,6 +83,27 @@
...
+        new MonetaryValue(0.0),
+        new MonetaryValue(-1.0),
+        new MonetaryValue(-2.0),
+        new MonetaryValue(-3.0),
+        new MonetaryValue(-4.0),
+        new MonetaryValue(-5.0),
+        new MonetaryValue(-6.0)
+      };
+
+  private final MonetaryValue[] BALANCE = {
```
alexanderkurash, line 96: BALANCE_POSITIVE?

src/test/java/org/folio/rest/impl/AccountsRefundAPITests.java
```diff
@@ -91,32 +112,36 @@ public void beforeEach() {
...
-    double transferAmount = 0.0;
-    double waiveAmount = 1.0;
-    double refundAmount = 5.0;
+    MonetaryValue initialAmount = new MonetaryValue(6.0);
+    MonetaryValue payAmount = new MonetaryValue(5.0);
+    MonetaryValue transferAmount = new MonetaryValue(0.0);
+    MonetaryValue waiveAmount = new MonetaryValue(1.0);
+    MonetaryValue refundAmount = new MonetaryValue(5.0);
+
+    MonetaryValue amountMinus5 = new MonetaryValue(-5.0);
```
alexanderkurash, line 121: What is the actual meaning of this -5 amount? If it's just negative refund amount, maybe we can use refundAmount instead?

src/test/java/org/folio/rest/impl/AccountsRefundAPITests.java
```diff
@@ -91,32 +112,36 @@ public void beforeEach() {
...
+    MonetaryValue amount0 = new MonetaryValue(0.0);
```
alexanderkurash, line 122: zeroAmount would probably be a better name

src/test/java/org/folio/rest/impl/AccountsRefundAPITests.java
```diff
@@ -125,17 +150,17 @@ public void fullRefundOfClosedAccountWithTransfer() {
...
+    MonetaryValue refundAmount = new MonetaryValue(5.0);
 
     List<Matcher<JsonObject>> expectedFeeFineActions = Arrays.asList(
-      feeFineActionMatcher(FIRST_ACCOUNT_ID, -3.0, payAmount, CREDIT.getFullResult(), REFUND_TO_PATRON),
-      feeFineActionMatcher(FIRST_ACCOUNT_ID, -5.0, transferAmount, CREDIT.getFullResult(), REFUND_TO_BURSAR),
-      feeFineActionMatcher(FIRST_ACCOUNT_ID, -2.0, payAmount, REFUND.getFullResult(), REFUNDED_TO_PATRON),
-      feeFineActionMatcher(FIRST_ACCOUNT_ID,  0.0, transferAmount, REFUND.getFullResult(), REFUNDED_TO_BURSAR)
+      feeFineActionMatcher(FIRST_ACCOUNT_ID, BALANCE_MINUS[3], payAmount, CREDIT.getFullResult(), REFUND_TO_PATRON),
+      feeFineActionMatcher(FIRST_ACCOUNT_ID, BALANCE_MINUS[5], transferAmount, CREDIT.getFullResult(), REFUND_TO_BURSAR),
+      feeFineActionMatcher(FIRST_ACCOUNT_ID, BALANCE_MINUS[2], payAmount, REFUND.getFullResult(), REFUNDED_TO_PATRON),
```
alexanderkurash, line 162: Again, if these values are just "reverted" payment, refund, transfer amount etc., maybe we can use the existing variables instead of creating new ones?

alexanderkurash:
Why do we want to merge this in b15.9 branch instead of master?

alexanderkurash (changes requested):
This should be a `feature branch` -> `master` PR

PR #195, opened 2021-08-04
[MODFEE-203] Consider previously refunded amount when repeatedly refunding the same fee/fine

OleksandrVidinieiev:
Resolves [MODFEE-203](https://issues.folio.org/browse/MODFEE-203)

### DIFF
--- a/src/test/java/org/folio/test/support/ApiTests.java
+++ b/src/test/java/org/folio/test/support/ApiTests.java
@@ -177,7 +177,7 @@ public class ApiTests {
 
   protected static <T> T get(CompletableFuture<T> future) {
     try {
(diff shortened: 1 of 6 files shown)


src/test/java/org/folio/test/support/ApiTests.java
```diff
@@ -177,7 +177,7 @@ protected static String randomId() {
 
   protected static <T> T get(CompletableFuture<T> future) {
     try {
-      return future.get(20, TimeUnit.SECONDS);
+      return future.get(60, TimeUnit.SECONDS);
```
OleksandrVidinieiev, line 180: It takes more that 20 seconds on my machine

alexanderkurash (approved):

PR #199, opened 2021-09-14
MODFEE-198 Add Contributors field to account

KotlyarovV:
https://issues.folio.org/browse/MODFEE-198
adding contributors field to account model


KotlyarovV:
https://issues.folio.org/browse/MODFEE-198 
Adding field "contributors" to account.

alexanderkurash (approved):

PR #206, opened 2021-12-15
MODFEE-232 Added log4j dependencies (version 2.16.0) into pom.xml

sherzodn:
https://issues.folio.org/browse/MODFEE-232

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -25,6 +25,7 @@
     <postgres.port>5433</postgres.port>
     <sonar.exclusions>**/impl/**</sonar.exclusions>
     <junit.jupiter.version>5.5.1</junit.jupiter.version>
+    <log4j2.version>2.16.0</log4j2.version>
   </properties>
 
   <dependencyManagement>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -17,14 +17,14 @@
     <powermock.version>2.0.7</powermock.version>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-    <raml-module-builder.version>33.0.0</raml-module-builder.version>
-    <vertx.version>4.1.0.CR1</vertx.version>
+    <raml-module-builder.version>33.2.2</raml-module-builder.version>
```
alexanderkurash, line 20: We don't need to upgrade to a newer minor version of RMB, just to the latest bugfix version - 33.0.x

alexanderkurash (changes requested):

alexanderkurash (approved):

PR #213, opened 2022-01-27
MODFEE-235 - Property is added: dateClosed

ccruz17:
### Story
Only the property is added: dateClosed

https://issues.folio.org/browse/MODFEE-235


alexanderkurash (changes requested):
This PR shouldn't be merged until backend processes that close fees/fines are aware of this field. Reach out to Holly Mistlebauer for details.

ccruz17:
Hi @alexanderkurash, 

Holly updated the issue details for the backend, I proceeded to submit new commits to resolve what Holly raised. Can it be rechecked? 
Thank you

cc: @zburke

src/main/java/org/folio/rest/service/action/RefundActionService.java
```diff
@@ -164,6 +164,7 @@ private static void updateAccountInMemory(Account account, Feefineaction feeFine
     account.setRemaining(feeFineAction.getBalance());
     account.getPaymentStatus().setName(fromValue(feeFineAction.getTypeAction()));
     account.getStatus().setName(OPEN.getValue());
+    account.setDateClosed(null);
```
alexanderkurash, line 167: Can `dateClose` change here from some non-null value to null? If it can, is there a test for this scenario?

ccruz17: Yes, there is already a test that goes through this code.

I just added the "Account" status validation, where it causes the close date to change to null ([commit](https://github.com/folio-org/mod-feesfines/pull/213/commits/986d4122b1b215023b91ffba63cddf683ba19454))

ccruz17 (commented):
@alexanderkurash
 I answered your questions and sent confirmations that resolve the details you mention.

ccruz17 (commented):
@alexanderkurash
 I have already applied the requested changes

ccruz17 (commented):
@alexanderkurash, I think the pull request is ready or something is missing that I am not aware of

src/main/java/org/folio/rest/service/AccountEventPublisher.java
```diff
@@ -37,7 +38,8 @@ public void publishAccountBalanceChangeEvent(Account account) {
   public void publishDeletedAccountBalanceChangeEvent(String accountId) {
     final Account account = new Account()
       .withId(accountId)
-      .withRemaining(new MonetaryValue(BigDecimal.ZERO));
+      .withRemaining(new MonetaryValue(BigDecimal.ZERO))
+      .withDateClosed(new Date());
```
Mykyta-Varenyk, line 42: Could you please explain, why do we even need to add the value here, if it is not used later in the event payload?

src/test/java/org/folio/rest/impl/AccountsAPITest.java
```diff
@@ -149,20 +156,26 @@ public void eventIsPublishedWhenFeeFineIsClosedWithLoanAndNoRemainingAmount() {
...
-      .put("remaining", 0.0);
+      .put("remaining", 0.0)
+      .put("dateClosed", DATE_CLOSED);
 
     accountsClient.update(accountId, updatedAccount);
 
     Response byId = accountsClient.getById(accountId);
     assertThat(byId, isPaidFully());
 
+    assertThat(accountsClient.getById(accountId).body().asString(), allOf(
```
Mykyta-Varenyk, line 175: please remove allOf and toString in assertions, they are not needed

src/test/java/org/folio/rest/impl/AccountsAPITest.java
```diff
@@ -149,20 +156,26 @@ public void eventIsPublishedWhenFeeFineIsClosedWithLoanAndNoRemainingAmount() {
     final JsonObject account = createAccountJsonObject(accountId)
       .put("loanId", loanId)
       .put("remaining", 90.00)
-      .put("status", createNamedObject("Open"));
+      .put("status", createNamedObject("Open"))
+      .put("dateClosed", null);
```
Mykyta-Varenyk, line 160: Could you please explain, why do we need to put a null here? I've run the test and added account get request by id after creation of this account, there was no dateClosed field

src/test/java/org/folio/rest/impl/AccountsAPITest.java
```diff
@@ -136,8 +143,8 @@ public void testAllMethodsAndEventPublishing() {
 
     Account accountToDelete = new Account()
       .withId(accountId)
-      .withRemaining(new MonetaryValue(new BigDecimal("0.00")));
-
+      .withRemaining(new MonetaryValue(new BigDecimal("0.00")))
+      .withDateClosed(new Date());
```
Mykyta-Varenyk, line 147: We are not checking date closed in assertion later. Please find a way to do that.

src/main/java/org/folio/rest/service/action/ActionService.java
```diff
@@ -157,8 +157,10 @@ protected Feefineaction createFeeFineActionAndUpdateAccount(Account account, Mon
     if (isFullAction) {
       account.getStatus().setName(CLOSED.getValue());
       account.setRemaining(MonetaryValue.ZERO);
+      account.setDateClosed(new Date());
     } else {
       account.setRemaining(feeFineAction.getBalance());
+      account.setDateClosed(null);
```
Mykyta-Varenyk, line 163: Please remove, as it's not doing anything

src/test/java/org/folio/test/support/ActionsAPITests.java
```diff
@@ -23,7 +24,8 @@ public static Account verifyAccountAndGet(ResourceClient accountsClient, String
       .body("remaining", is((float) amount.toDouble()))
       .body("status.name", is(statusName))
       .body("paymentStatus.name", is(expectedPaymentStatus))
-      .body("metadata.updatedDate", notNullValue());
+      .body("metadata.updatedDate", notNullValue())
+      .body("dateClosed", (statusName== "Open" ? nullValue() : notNullValue()));
```
Mykyta-Varenyk, line 28: Please check that dateClosed is correct

src/main/java/org/folio/rest/service/AccountEventPublisher.java
ccruz17: Actually at this point it is my mistake, since when deleting the Account, it does not require the closing date, because you are deleting, sorry

Submit a new commit to fix this https://github.com/folio-org/mod-feesfines/pull/213/commits/3e0d66752f02cfad3087c2b5653d7d8145fe68ef

src/test/java/org/folio/test/support/ActionsAPITests.java
ccruz17: At this point I can't check the exact date, as the date is updated and I lose the exact second. However, I checked the log and it works fine, that's why I only set it to whether it is null or not

src/test/java/org/folio/rest/impl/AccountsAPITest.java
ccruz17: If I update the buildAccount function and add withDateClose(null), which is called by the createAccountJsonObject function, the JSON never returns the dateClosed field since it is initially null. So just for testing I add it here to verify that the property exists and then assess that the correct value has been added

src/test/java/org/folio/rest/impl/AccountsAPITest.java
ccruz17: I only use the assertion to check if the closing date changed after the update. I don't think it's necessary to remove them.

ccruz17 (commented):
@Mykyta-Varenyk, Can you help me validate again, I have answered your questions

ccruz17 (commented):
The changes have been applied

PR #218, opened 2022-02-23
MODFEE-220: Make paymentStatus property in accountdata.json required and also make it an enum

yusuf-murodov-epam:
## Purpose
[MODFEE-220](https://issues.folio.org/browse/MODFEE-220)
_Make paymentStatus property in accountdata.json required and also make it an enum_

### DIFF
--- a/ramls/accountdata.json
+++ b/ramls/accountdata.json
@@ -38,17 +38,9 @@
       ]
     },
     "paymentStatus": {
-      "description": "Overall status of the payment/waive/transfer/refund/cancel",
-      "type": "object",
-      "properties": {
-        "name": {
-          "description": "Name of the status (values used are Outstanding, Paid partially, Paid fully, Waived partially, Waived fully, Transferred partially, Transferred fully, Refunded partially, Refunded fully, Cancelled as error)",
-          "type": "string"
-        }
-      },
-      "required": [
-        "name"
-      ]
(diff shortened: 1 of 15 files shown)


sherzodn (approved):

bisecomsergiy (approved):

ramls/accountdata.json
```diff
@@ -38,17 +38,9 @@
...
-        "name": {
-          "description": "Name of the status (values used are Outstanding, Paid partially, Paid fully, Waived partially, Waived fully, Transferred partially, Transferred fully, Refunded partially, Refunded fully, Cancelled as error)",
-          "type": "string"
-        }
-      },
-      "required": [
-        "name"
-      ]
+      "description": "Overall status of the payment/waive/transfer/refund/credit/cancel",
+      "type": "string",
```
alexanderkurash, line 42: We can't change the type from object to string. This is a breaking change - it will break the UI and all of the modules that create accounts. The reason is that previously this field was serialized to JSON like this:
```
"paymentStatus": {
  "name": "Paid fully"
}
```
but after your change it will look like this
```
"paymentStatus": "Paid fully"
```
you shouldn't change `paymentStatus`'es type from object to string, you should make a `name` field of the `paymentStatus` an enum.

alexanderkurash (changes requested):

PR #219, opened 2022-02-28
MODFEE-243 Prevent deleting/updating required FeeFines

sherzodn:
https://issues.folio.org/browse/MODFEE-243

### DIFF
--- a/src/main/java/org/folio/rest/impl/FeeFinesAPI.java
+++ b/src/main/java/org/folio/rest/impl/FeeFinesAPI.java
@@ -1,5 +1,8 @@
 package org.folio.rest.impl;
 
+import static org.folio.rest.jaxrs.resource.Feefines.PostFeefinesResponse.respond422WithApplicationJson;
+import static org.folio.rest.utils.ErrorHelper.createErrors;
+
 import java.io.IOException;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


Mykyta-Varenyk (approved):

bisecomsergiy (approved):

src/main/java/org/folio/rest/impl/FeeFinesAPI.java
```diff
@@ -42,6 +46,10 @@
     private static final String FEEFINE_ID_FIELD = "'id'";
     private static final String OKAPI_HEADER_TENANT = "x-okapi-tenant";
     private final Logger logger = LogManager.getLogger(FeeFinesAPI.class);
+    private final List<String> requiredFeeFineIdList = List.of("9523cb96-e752-40c2-89da-60f3961a488d",
```
alexanderkurash, line 49: We call these fee/fine types automatic - check populate-feefines.sql
Also, should we make an enum for these values?

sherzodn: so, you propose to create an enum for these ID values, right?

alexanderkurash: Yes. It would be nice to have a single place in the code with this list, otherwise we'll be copy/pasting these values everywhere. You can make it similar to patron block conditions in mod-patron-blocks

sherzodn: done

PR #221, opened 2022-03-16
MODFEE-243 - Prevent deleting/updating required FeeFines

sherzodn:
https://issues.folio.org/browse/MODFEE-243

### DIFF
--- a/src/test/java/org/folio/rest/impl/FeeFinesAPITest.java
+++ b/src/test/java/org/folio/rest/impl/FeeFinesAPITest.java
@@ -1,5 +1,7 @@
 package org.folio.rest.impl;
 
+import static org.hamcrest.CoreMatchers.is;
+
 import org.apache.http.HttpStatus;
 import org.folio.rest.domain.AutomaticFeeFineType;
 import org.folio.test.support.ApiTests;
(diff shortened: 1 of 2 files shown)


src/test/java/org/folio/rest/impl/FeeFinesAPITest.java
```diff
@@ -16,14 +17,6 @@
...
 public class FeeFinesAPITest extends ApiTests {
   private static final String REST_PATH = "/feefines";
-  private static final String FEEFINES_TABLE = "feefines";
-  private static final String OWNERS_TABLE = "owners";
-
-  @Before
-  public void setUp() {
-    removeAllFromTable(FEEFINES_TABLE);
-    removeAllFromTable(OWNERS_TABLE);
-  }
```
alexanderkurash, line 26: Why did we remove these lines? I guess they delete automatic fees/fines, but they were added here to guarantee that tests are independent. I would rather make one of them to delete everything except automatic types.

sherzodn: fixed

src/test/java/org/folio/test/support/ApiTests.java
```diff
@@ -155,13 +156,25 @@ protected OkapiDeployment getOkapi() {
...
+      }
+    }
     PostgresClient.getInstance(vertx, TENANT_NAME)
-      .delete(tableName, new Criterion(), result -> future.complete(null));
+      .delete(tableName, criterion, result -> future.complete(null));
 
     get(future);
   }
 
+  private Criteria createCriteria(AutomaticFeeFineType automaticFeeFineType){
```
alexanderkurash, line 171: Missing space. Also, let's give it a more specific name. Like "createAutomaticFeeFineExclusionCriteria"

src/test/java/org/folio/test/support/ApiTests.java
```diff
@@ -155,13 +156,25 @@ protected OkapiDeployment getOkapi() {
 
   protected void removeAllFromTable(String tableName) {
     final CompletableFuture<Void> future = new CompletableFuture<>();
-
+    Criterion criterion = new Criterion();
+    if ("feefines".equals(tableName)) {
```
alexanderkurash, line 160: What if we move the constants with table names from FeeFinesAPITest to ApiTests and use it here

alexanderkurash (approved):

PR #228, opened 2022-04-21
MODFEE-244 Reformat code

Mykyta-Varenyk:


src/main/java/org/folio/rest/service/report/utils/LookupHelper.java
```diff
@@ -223,7 +223,7 @@ public <T extends HasAccountInfo> Future<T> lookupRefundPayTransferFeeFineAction
     }
 
     return feeFineActionRepository.findActionsOfTypesForAccount(accountId,
-      List.of(REFUND, PAY, TRANSFER))
+        List.of(REFUND, PAY, TRANSFER))
       .map(this::sortFeeFineActionsByDate)
       .map(actions -> ctx.updateAccountContextWithActions(accountId, actions))
       .map(ctx)
```
alexanderkurash, line 229: Indentation?

src/main/java/org/folio/rest/service/report/context/HasAccountInfo.java
```diff
@@ -9,7 +9,10 @@
 
 public interface HasAccountInfo {
   Account getAccountById(String accountId);
+
   Future<Void> updateAccountContextWithActions(String accountId, List<Feefineaction> actions);
+
   boolean isAccountContextCreated(String accountId);
+
```
alexanderkurash, line 16: Do we always put new lines between fields in classes?

Mykyta-Varenyk: We do have lines between methods

src/main/java/org/folio/rest/service/report/utils/LookupHelper.java
Mykyta-Varenyk: Yes, this is a method parameter and is not related to the stream operations below

PR #230, opened 2022-05-12
MODFEE-253 Lost item processing fee can be saved with a value less than 0

bisecomsergiy:
### Purpose
Avoid entering values less then 0 to Lost item processing fee 
### Approach
Add validation
Add tests
### Learning
https://issues.folio.org/browse/MODFEE-253

### DIFF
--- a/src/main/java/org/folio/rest/impl/LostItemFeePoliciesAPI.java
+++ b/src/main/java/org/folio/rest/impl/LostItemFeePoliciesAPI.java
@@ -5,14 +5,17 @@ import io.vertx.core.Context;
 import io.vertx.core.Handler;
 
 import java.util.Map;
+import java.util.Optional;
 
 import javax.ws.rs.core.Response;
 
 import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.folio.rest.annotations.Validate;
+import org.folio.rest.domain.MonetaryValue;
 import org.folio.rest.jaxrs.model.LostItemFeePolicy;
 import org.folio.rest.jaxrs.model.LostItemFeesPoliciesGetOrder;
 import org.folio.rest.jaxrs.model.LostItemFeePolicies;
+import org.folio.rest.jaxrs.model.OverdueFine;
 import org.folio.rest.jaxrs.resource.LostItemFeesPolicies;
 import org.folio.rest.persist.PgUtil;
 
@@ -122,7 +125,10 @@ public class LostItemFeePoliciesAPI implements LostItemFeesPolicies {
   }
 
   private boolean containsNegativeValue(LostItemFeePolicy entity) {
-    return entity.getLostItemProcessingFee().isNegative();
+    return Optional.ofNullable(entity)
+      .map(LostItemFeePolicy::getLostItemProcessingFee)
+      .map(MonetaryValue::isNegative)
+      .orElse(false);
   }
 
   private String getAmountValueAsString(LostItemFeePolicy entity) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/folio/rest/impl/LostItemFeePoliciesAPI.java
```diff
@@ -94,7 +105,33 @@ public void putLostItemFeesPoliciesByLostItemFeePolicyId(String lostItemFeePolic
...
+
     PgUtil.put(TABLE_NAME, entity, lostItemFeePolicyId, okapiHeaders, vertxContext,
       PutLostItemFeesPoliciesByLostItemFeePolicyIdResponse.class, asyncResultHandler);
   }
+
+  private void handleError(LostItemFeePolicy entity,
+    Handler<AsyncResult<Response>> asyncResultHandler) {
+
+      ImmutablePair<String, String> errorInformation = new ImmutablePair<>("lostItemProcessingFee",
+        getAmountValueAsString(entity));
```
alexanderkurash, line 121: Why do we need this pair? Can these values be inlined?

src/main/java/org/folio/rest/impl/LostItemFeePoliciesAPI.java
```diff
@@ -94,7 +103,28 @@ public void putLostItemFeesPoliciesByLostItemFeePolicyId(String lostItemFeePolic
...
+        succeededFuture(respond422WithApplicationJson(createError(
+          NEGATIVE_VALUE_MESSAGE, "lostItemProcessingFee",
+          entity.getLostItemProcessingFee().getAmount().toString()))));
+  }
+
+  private boolean shouldRefuseWhenLostItemProcessingFeeNegative(LostItemFeePolicy entity) {
+    return Optional.ofNullable(entity)
+      .map(LostItemFeePolicy::getLostItemProcessingFee)
+      .map(MonetaryValue::isNegative)
+      .orElse(false);
```
alexanderkurash, line 128: Make 1 method that returns boolean and handles the error

src/main/java/org/folio/rest/impl/LostItemFeePoliciesAPI.java
bisecomsergiy: It is refactored, thanks!

src/main/java/org/folio/rest/impl/LostItemFeePoliciesAPI.java
```diff
@@ -94,7 +105,33 @@ public void putLostItemFeesPoliciesByLostItemFeePolicyId(String lostItemFeePolic
...
+    if (containsNegativeValue(entity)) {
+      handleError(entity, asyncResultHandler);
+      return;
+    }
+
     PgUtil.put(TABLE_NAME, entity, lostItemFeePolicyId, okapiHeaders, vertxContext,
       PutLostItemFeesPoliciesByLostItemFeePolicyIdResponse.class, asyncResultHandler);
   }
+
+  private void handleError(LostItemFeePolicy entity,
```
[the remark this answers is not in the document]
bisecomsergiy: The method is removed

src/main/java/org/folio/rest/impl/LostItemFeePoliciesAPI.java
bisecomsergiy: The methods are coupled, thanks.

alexanderkurash (approved):

PR #237, opened 2022-06-23
MODFEE-262 Make Lost item fee (actual cost) fee/fine type read-only

viktor-draban:
[ticket](https://issues.folio.org/browse/MODFEE-262)

Make Lost item fee (actual cost) fee/fine type read-only


src/main/java/org/folio/rest/domain/AutomaticFeeFineType.java
```diff
@@ -9,6 +9,7 @@ public enum AutomaticFeeFineType {
   OVERDUE_FINE("9523cb96-e752-40c2-89da-60f3961a488d"),
   REPLACEMENT_PROCESSING_FEE("d20df2fb-45fd-4184-b238-0d25747ffdd9"),
   LOST_ITEM_FEE("cf238f9f-7018-47b7-b815-bb2db798e19f"),
+  LOST_ITEM_FEE_ACTUAL_COST("73785370-d3bd-4d92-942d-ae2268e02ded"),
```
roman-barannyk, line 12:
```suggestion
  LOST_ITEM_ACTUAL_COST_FEE("73785370-d3bd-4d92-942d-ae2268e02ded"),
```

PR #267, opened 2023-02-23
MODFEE-315 Fee/fine action migration

OleksandrVidinieiev:
Resolves [MODFEE-315](https://issues.folio.org/browse/MODFEE-315)

**Approach**
Update 'createdAt' in every fee/fine action where this property is non-null and has non-UUID value.
Candidates for new value in order of preference:
- if current value is a service point name, find its ID
- if user associated with this action has a default service point set up, find its ID
- fallback service point ID provided by tenant


OleksandrVidinieiev:
This solution was rejected in favor of #268

PR #268, opened 2023-03-10
MODFEE-315 Enable UUID-validation for 'createdAt', backup invalid values

OleksandrVidinieiev:
- only allow UUIDs in property `createdAt` of fee/fine action
- copy existing invalid (non-UUID) values into new field `originalCreatedAt`, remove `createdAt` key from affected actions afterwards

Resolves [MODFEE-315](https://issues.folio.org/browse/MODFEE-315)


src/test/java/org/folio/test/support/ApiTests.java
```diff
@@ -148,7 +150,7 @@ protected static TenantAttributes getTenantAttributes() {
 
     return new TenantAttributes()
       .withModuleFrom(MODULE_NAME + "-14.2.4")
-      .withModuleTo(MODULE_NAME + "-" + getModuleVersion())
+      .withModuleTo(PomUtils.getModuleId())
```
OleksandrVidinieiev, line 153: Unlike `getModuleId`, method `getModuleVersion()` returns current module version with `*-SNAPSHOT` suffix, which results in migration scripts running regardless of `fromModuleVersion` specified in the schema

PR #272, opened 2023-03-24
MODFEE-328 Add the field "effectiveLocationDiscoveryDisplayName" to the Item Object in the Notice context

pavankumar181:
**Purpose**
This PR serves the purpose to add the value of Item.effectiveLocationDiscoveryDisplayName in the Notice Context which is required as per [https://issues.folio.org/browse/MODFEE-328](url) for Manual fee/fine charge + Manual fee/fine action (pay, waive, refund, transfer or cancel/error).
**Approach**
We already have the field discoveryDisplayName in Location Object so at the time of preparing Notice context
we are adding the above value and set it to the Item Context.
After this change, LocationDiscoveryDisplayName is populated in the Notices. Please find the Email Evidence below
MODFEE-328-2
MODFEE-328-1

### DIFF
--- a/src/main/java/org/folio/rest/utils/PatronNoticeBuilder.java
+++ b/src/main/java/org/folio/rest/utils/PatronNoticeBuilder.java
@@ -123,6 +123,7 @@ public class PatronNoticeBuilder {
 
     if (location != null) {
       itemContext.put(EFFECTIVE_LOCATION_SPECIFIC, location.getName());
+      itemContext.put("effectiveLocationDiscoveryDisplayName", location.getDiscoveryDisplayName());
 
       if (location.getLibrary() != null) {
         itemContext.put("effectiveLocationLibrary",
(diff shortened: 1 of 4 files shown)


src/main/java/org/folio/rest/utils/PatronNoticeBuilder.java
gurleenkaurbp, line 1: Please create a constant for effectiveLocationDiscoveryDisplayName like EFFECTIVE_LOCATION_SPECIFIC

gurleenkaurbp (changes requested):

OleksandrVidinieiev (approved):

src/main/java/org/folio/rest/utils/PatronNoticeBuilder.java
pavankumar181: completed

Vignesh-kalyanasundaram (approved):

gurleenkaurbp (approved):

PR #278, opened 2023-06-19
MODFEE-356

nielserik:


nielserik:
This PR is about six additional properties and an extra enum value in the new reminder fees policy sub-section of the overdue fines policy schema. 

There are no functional changes at this stage.

PR #299, opened 2024-11-01
[MODFEE-385] - Review and cleanup Module Descriptors for mod-feesfines

azizbekxm:
## Purpose
https://folio-org.atlassian.net/browse/MODFEE-385

## Approach
- adding permission module name as prefix
- fixing incorrect permissions

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -54,7 +54,7 @@
   "provides":[
     {
       "id":"feesfines",
-      "version": "19.0",
+      "version": "19.1",
       "handlers":[
         {
           "methods":[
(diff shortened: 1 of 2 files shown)


descriptors/ModuleDescriptor-template.json
```diff
@@ -208,7 +208,7 @@
           ],
           "pathPattern":"/accounts/{id}/check-pay",
           "permissionsRequired":[
-            "accounts.check-pay.post"
+            "feesfines.accounts.check-pay.item.post"
```
OleksandrVidinieiev, line 211: This endpoint represents an action performed on "account"(s). And these actions are not "items", not really. So a more appropriate permission name in this case would be `feesfines.accounts.check-pay.post` or `feesfines.accounts.check-pay.execute`. Same applies to all other action-related permissions below.

azizbekxm: Thanks, fixed

alexanderkurash (approved):

PR #312, opened 2025-03-24
MODFEE-409 add tokens to item context for manual charges/actions

nielserik:
Bug-fix for Sunflower.

The PR suggests that the changes that are raised for merge to master in PR #311 are also applied to the release branch for release 19.3, after which version 19.3.1 could be released with this bug-fix.  

Add six new tokens (and support a seventh) to the item context for use in relation to reading room circulation (in-house use requests).

item.accessionNumber
item.administrativeNotes (from instance.adminstrativeNotes)
item.datesOfPublication (from instance.publication)
item.editions (from instance.editions)
item.physicalDescriptions (from instance.physicalDescriptions)
item.instanceHrid

The addition of instance HRID will support rendering an instance HRID barcode image.


src/main/java/org/folio/rest/utils/PatronNoticeBuilder.java
```diff
@@ -235,6 +243,12 @@ private static String getAllContributors(Instance instance) {
       .collect(joining(LIST_VALUES_SEPARATOR));
   }
 
+  private static String getDatesOfPublication(Instance instance) {
+    return instance.getPublication().stream()
```
OleksandrVidinieiev, line 247: Will return value of `instance.getPublication()` always be non-null?

src/test/java/org/folio/rest/impl/FeeFineActionsAPITest.java
```diff
@@ -338,12 +338,18 @@ public void postActionWithPatronNotice() {
...
+          .put("accessionNumber", item.getAccessionNumber())
           .put("callNumber", item.getEffectiveCallNumberComponents().getCallNumber())
           .put("callNumberPrefix", item.getEffectiveCallNumberComponents().getPrefix())
           .put("callNumberSuffix", item.getEffectiveCallNumberComponents().getSuffix())
           .put("title", instance.getTitle())
+          .put("instanceHrid", instance.getHrid())
           .put("primaryContributor", PRIMARY_CONTRIBUTOR_NAME)
           .put("allContributors", PRIMARY_CONTRIBUTOR_NAME + "; " + NON_PRIMARY_CONTRIBUTOR_NAME)
+          .put("physicalDescriptions", "Hardback; Green")
+          .put("editions", "1st; 2nd")
```
OleksandrVidinieiev, line 350: As far as I remember, array of editions is deserialized into a set, which means the order of its elements may change, so you can end up with "1st; 2nd" or "2nd; 1st".

src/main/java/org/folio/rest/utils/PatronNoticeBuilder.java
nielserik: Yes.

src/test/java/org/folio/rest/impl/FeeFineActionsAPITest.java
nielserik: All right, testing with list of one.

PR #313, opened 2025-03-25
MODFEE-409 add tokens to item context for manual charges/actions

nielserik:


Jenkinsfile
```diff
@@ -1,3 +1,5 @@
+@Library('folio_jenkins_shared_libs@test-sonar-maven-update') _
```
OleksandrVidinieiev, line 1: What is this?

funkymalc: just some Sonar maven plugin testing/troubleshooting @OleksandrVidinieiev

nielserik: John is looking into the test issues that I asked about in the DM chat between the two of us Monday, @OleksandrVidinieiev

